Issue #172, opened 2022-08-09
Jupyter notebook for running scripts


thusser:
See LSST

PR #263, opened 2023-09-13
Added tutorial for jupyter notebooks

GermanHydrogen:
Added a tutorial for using pyobs in jupyter notebooks.


GermanHydrogen:
Probably closes #172

PR #328, opened 2023-12-27
Added exception handler image processor decorator

GermanHydrogen:
This is an alternative suggestion to implementing the exception handling in the image processor base class.
Doing this in the base class would introduce two new attributes `suppress_exception: bool` which toggles the exception handling and `error_header: Optional[str]` which defines the error header flat. This introduces the problem that the `error_header` attribute can be set without the `suppress_exception` attribute being true, which could lead to errors in config management. 
The alternative decorator approach removes the need for the first attribute.
A config with this could look like this

```yaml

[...]
pipeline:

  - class: pyobs.images.processors.detection.SepSourceDetection
  - class: pyobs.image.processors.ExceptionHandler
      processor:
        class: pyobs.images.processors.astrometry.AstrometryDotNet
          url: ...
          radius: 5
      error_header: "WCSERR"
[...]

```

I'm open to feedback.

### DIFF
--- a/pyobs/images/processors/__init__.py
+++ b/pyobs/images/processors/__init__.py
@@ -2,3 +2,7 @@
 Image Processors (pyobs.images.processors)
 ------------------------------------------
 """
+
+from .exceptionhandler import (ExceptionHandler)
+
+__all__ = ["ExceptionHandler"]
--- a/pyobs/images/processors/exceptionhandler.py
+++ b/pyobs/images/processors/exceptionhandler.py
@@ -1,0 +1,41 @@
+import logging
+from typing import Union, Dict, Any, Optional
+
+from pyobs.images import ImageProcessor, Image
+from pyobs.object import get_object
+
+import pyobs.utils.exceptions as exc
+log = logging.getLogger(__name__)
+
+
+class ExceptionHandler(ImageProcessor):
+
+    __module__ = "pyobs.images.processors"
+
+    def __init__(self, processor: Union[ImageProcessor, Dict[str, Any]], error_header: Optional[str] = None) -> None:
+        super().__init__()
+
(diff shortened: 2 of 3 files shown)


thusser:
Hmm, not a bad idea... But that moves error handling in the config file, which I find weird... The config also gets a lot more complex with this. What about adding that functionality to the `ImageProcessor` base class?

thusser:
Revisiting this: the actual problem is that `PipelineMixin.run_pipeline()` aborts the whole chain and propagates on the first exception from *any* step, with no way to make one step's failure non-fatal (log and continue) while another step in the same pipeline stays fatal (raise and abort the run).

Given the objection above was about config complexity from wrapping steps in a separate `ExceptionHandler` node, an alternative shape: add an optional kwarg directly on `ImageProcessor`'s base `__init__` (e.g. `on_error: "raise" | "log"`, default `"raise"`) instead of a wrapper class. Since every processor already accepts arbitrary config kwargs via `Object`, this is one extra line on the step's own config rather than nesting it inside a decorator processor:

```yaml
pipeline:
  - class: pyobs.images.processors.detection.SepSourceDetection
  - class: pyobs.images.processors.astrometry.AstrometryDotNet
    url: ...
    radius: 5
    on_error: log
```

`PipelineMixin.run_pipeline()`'s loop would then catch per-step and check `step.on_error` to decide whether to log-and-continue (image passes through unmodified for that step) or let the exception propagate as it does today. Smaller diff than this PR's decorator approach, and keeps the toggle next to the step it affects rather than in a separate wrapper layer -- but doesn't get you a custom error-header tag like `error_header` did here; that'd need to be a second optional kwarg alongside `on_error` if still wanted.

Happy to implement this version if there's still interest in solving it.

thusser:
Closing in favor of #693, which captures the underlying need without this PR's config-nesting approach. Thanks for raising the idea originally!

PR #338, opened 2024-01-23
NStarOffsets testing & refactor

GermanHydrogen:
Tasks:
- [x] Tests
- [x] Refactoring

Fixed:
- Wrong default value if Gaussian fit fails
https://github.com/pyobs/pyobs-core/blob/d5f79525842a7ff8cad0e650dd10b867385865de/pyobs/images/processors/offsets/nstar.py#L359

Problems:
- Box-Size is capped at 5x5 pixels which seems to small -> Change max_offset to pixel
https://github.com/pyobs/pyobs-core/blob/9d0feb85ca8a6ccbd788a30c73731df9a2f447ab/pyobs/images/processors/offsets/nstar.py#L80
- Default Box-Size is incompatible with `photutils.psf.extract_stars`, which expects odd box-sizes
https://github.com/pyobs/pyobs-core/blob/9d0feb85ca8a6ccbd788a30c73731df9a2f447ab/pyobs/images/processors/offsets/nstar.py#L111
- The mean of the box offsets seems to be a bad choice for a location parameter as it is not robust. The median or $\alpha$-trimmed mean would be a better choice. -> Different cases for n
https://github.com/pyobs/pyobs-core/blob/9d0feb85ca8a6ccbd788a30c73731df9a2f447ab/pyobs/images/processors/offsets/nstar.py#L297
- Unconventional background contrast calculation.  -> Extract to filter
https://github.com/pyobs/pyobs-core/blob/9d0feb85ca8a6ccbd788a30c73731df9a2f447ab/pyobs/images/processors/offsets/nstar.py#L218
-  median squared relative residue (?) as a metric for fit quality
https://github.com/pyobs/pyobs-core/blob/9d0feb85ca8a6ccbd788a30c73731df9a2f447ab/pyobs/images/processors/offsets/nstar.py#L366C2-L366C48


GermanHydrogen:
- Changed the unit of the  class parameter `max_offset` to pixel
- Seperated the source filter from the box generator and created a new processor `ImageSourceFilter`
- Changed the background contrast metric to weber contrast

After further thought, I did not change the location parameter for calculating the offset from the box offsets.
I also did not change the fit quality metric. Both things should be first tested under real conditions, to determine if they are sufficient.

PR #345, opened 2024-02-12
Telescope Test and Refactor

GermanHydrogen:
## TODO
- [x] Unit test `BaseTelescope`
- [ ] Refactor `BaseTelescope`
- [x] Unit test `DummyTelescope`
- [ ] Refactor `DummyTelescope`

### DIFF
--- a/pyobs/modules/telescope/basetelescope.py
+++ b/pyobs/modules/telescope/basetelescope.py
@@ -206,67 +206,65 @@ class BaseTelescope(
         # define base header
         hdr: Dict[str, Union[Any, Tuple[Any, str]]] = {}
 
-        # positions
+        await self._add_coordinate_header(hdr)
+        await self._add_side_information_header(hdr)
+
+        self._copy_header(self._fits_headers, hdr)
+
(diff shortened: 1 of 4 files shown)


GermanHydrogen:
@thusser Do you know, why we are updating the "celestial" header every 30sec? 
https://github.com/pyobs/pyobs-core/blob/d5f79525842a7ff8cad0e650dd10b867385865de/pyobs/modules/telescope/basetelescope.py#L261

I do not see any reason, why this can't be calculated, when `get_fits_header_before` is called. Even if it would be computationally expensive, caching the result for 30sec seems like the better approach.

thusser:
Closing as stale: this targets pyobs/modules/telescope/dummytelescope.py, which no longer exists -- it was split into dummyaltaztelescope.py/dummyradectelescope.py/dummysolartelescope.py + _dummytelescopebase.py a while back, so this no longer applies without a full rewrite against the current module structure. Also has real merge conflicts against develop now (old Optional[Dict[...]] typing vs. the dict[...] | None style used throughout today).

Happy to revisit the underlying refactor/test-coverage goal against the current structure if there's still interest.

PR #357, opened 2024-04-11
Derotator Offsets

LeonMee:
Added a function that calculates the derotator offset based on a head…er and a given observation time.

### DIFF
--- a/pyobs/modules/telescope/basetelescope.py
+++ b/pyobs/modules/telescope/basetelescope.py
@@ -1,7 +1,9 @@
 import asyncio
 from abc import ABCMeta, abstractmethod
 from typing import Dict, Any, Tuple, Union, List, Optional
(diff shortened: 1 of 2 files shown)


pyobs/modules/telescope/basetelescope.py
```diff
@@ -308,5 +310,12 @@ async def _update_celestial_headers(self) -> None:
             "SUNDIST": (None if sun_dist is None else float(sun_dist.degree), "Solar Distance from Target"),
         }
 
+    def _get_derotator_offset_from_header(self, hdr, obstime):
```
thusser, line 313: Can we rename this to `_calculate_derotator_position_from_header`?

thusser: Or better: `_calculate_derotator_position` and don't use the header. The telescope itself knows best, where it's pointing...

PR #761, opened 2026-08-16
Explicit pubsub event subscriptions, decoupled from PEP presence

thusser:
## Summary

- Events now publish/subscribe via the shared `pubsub.<domain>` service using explicit XEP-0060
  subscribe/unsubscribe, mirroring the mechanism already used for module state
  (`_subscribe_state`/`_subscribe_with_retry`/`_unsubscribe_state`), instead of XEP-0163 PEP +
  presence auto-subscribe.
- A module now only receives event types it actually registered a handler for. Previously every
  module received and parsed every event in the fleet regardless of interest, filtering only
  after the fact at the application layer (`_send_event_to_module`).
- `add_interest()` / XEP-0163 PEP are no longer used for events at all; the `xep_0163` plugin
  registration is removed from `XmppClient`.
- Subscriptions are further gated on the peer's disco#info `role="send"` advertisement
  (`_event_role`/`_get_disco_info`): a module only subscribes to a peer's event node if that peer
  actually declares it sends that event type/version. Previously every module subscribed to
  every online peer for every event type it handled, regardless of whether the peer ever
  published it -- e.g. a camera's `BadWeatherEvent` handler would retry-subscribe to
  `admin:BadWeatherEvent` forever, spamming "still failing to subscribe" on a node that would
  never be created.
- No ejabberd server config change needed on any site -- this reuses infrastructure state
  delivery already depends on.

See `specs/adrs/0012-event-delivery-explicit-pubsub-subscription-not-presence.md` and
`specs/plans/2026-08-16-explicit-pubsub-event-subscriptions.md` for the full design discussion,
including why the originally-considered PEP-`access_model` approach was rejected (depended on
unverified ejabberd `node_pep` internals). Background:
`logevent-double-delivery-fix-discussion.md`.

Rollout is a coordinated cutover per site (every module restarts onto the new release together;
old PEP-publishing and new pubsub-subscribing modules cannot interoperate) -- see the plan doc's
rollout section. Not yet rolled out to any production site.

## Test plan

- [x] `tests/integration/test_xmpp_event_subscriptions.py` (9 tests) against the live
      `tests/xmpp/docker-compose.yml` harness: handler receives events from a peer; a module with
      no handler never receives the wire message at all (not just "receives and drops"); a
      handler registered after a peer is already online still subscribes; unregistering a
      handler stops further delivery; a handler for a local event (`ModuleOpenedEvent`/
      `ModuleClosedEvent`) never attempts a subscribe; `ModuleOpenedEvent`/`ModuleClosedEvent`
      still fire normally; a restarted session re-subscribes to an already-online peer; a peer
      that never advertises `role="send"` for an event type is never subscribed to (verified
      this fails without the gating fix); a peer that declares an event handler-less
      (send-only) still gets subscribed to and delivers normally.
- [x] Full existing XMPP integration suite (other `test_xmpp_*.py` files): 30/30 passing (state,
      presence, RPC, ACL, dummy camera untouched).
- [x] Full non-integration unit suite: 1452 passed (pre-existing, unrelated `test_pyobsd.py`
      failures confirmed present on `develop` too).
- [x] `ruff`, `pyrefly`, `black` clean on all changed files.


thusser (commented):
Two bugs, two lower-severity issues, and doc-consistency nits.

## Blocking

**1. `_got_online` over-subscribes, spinning forever on never-created nodes.**
The new block iterates `for ev in self._event_handlers`, which includes (a) local events `ModuleOpenedEvent`/`ModuleClosedEvent` (every module registers handlers for both via `module.py:338` and `comm.py:86`, but nobody ever publishes to `pyobs:event:<peer>:ModuleOpenedEvent:1`), and (b) event classes whose last handler was removed. On unregister, `unregister_event()` (`comm.py:473-478`) removes the handler and discards from `_events_subscribed` but never deletes the `_event_handlers` key, so an empty-list key remains. Result: for every peer coming online, each module spawns `_subscribe_event_with_retry` tasks against nodes that will never exist, retrying forever with capped backoff and logging "Still failing to subscribe…" every ~30 attempts (2×N permanent background tasks per module).

Fix the gate, e.g.:
```python
for ev, handlers in self._event_handlers.items():
    if ev.local or not handlers:
        continue
    asyncio.create_task(self._subscribe_event_with_retry(module_name, ev))
```

**2. `send_event` dereferences `self._module.name` unguarded.**
The node computation runs before the method's own `if self._module is not None` local-delivery guard, so a module-less `XmppComm` (GUI, admin tool, observer) emitting a `LogEvent` via `_logging` (`comm.py:376`) raises `AttributeError`, which `_logging`'s `except Exception` swallows and turns into repeated "Something went wrong" spam while the event is dropped. Event publishing previously needed no module identity. Guard it with a fallback (e.g. `self.client.boundjid.username`).

## Lower severity

**3. `_unregister_events` only unsubscribes from currently-online peers.** If the last handler is removed while a peer is offline (or races offline), the `(peer, ev)` key survives in `_event_subscriptions` and the server-side subscription stays, so the module keeps receiving that event's wire message (dropped at app layer). Discard keys for all known subscriptions, not just online peers.

**4. Test gap vs. the plan's stated validation.** The plan lists 6 steps but the PR ships 4 tests; the restart/re-subscribe case (step 5) and ModuleOpened/Closed sanity (step 6) claim "confirmed in testing" with no test backing them. More importantly, no test registers a local-event handler, so bug #1 is uncaught. Add a test asserting a module registering `ModuleOpenedEvent` does not attempt a subscribe to a peer's never-created node.

## Nits

- ADR 0012 is `status: proposed` but merges with the implementation (10 of 12 ADRs use `accepted`); plan body says "implemented" while `specs/plans/index.md` lists it "proposed".
- `logevent-double-delivery-fix-discussion.md` sits at repo root, outside `specs/` where CLAUDE.md says design notes live.

thusser:
Addressed in 68e71815:

**Blocking**
1. `_got_online`'s subscribe loop now filters `self._event_handlers.items()` to `not ev.local and handlers` — skips `ModuleOpenedEvent`/`ModuleClosedEvent` and stale empty-handler-list keys left by `unregister_event()`. Added `test_local_event_handler_never_subscribes` (registers a `ModuleOpenedEvent` handler, asserts `_event_subscriptions` stays empty after a peer comes online) — confirmed this fails without the fix.
2. `send_event()` now computes `publisher = self._module.name if self._module is not None else self.client.boundjid.user` before building the node id, instead of dereferencing `self._module.name` unconditionally ahead of the existing `if self._module is not None` guard.

**Lower severity**
3. `_unregister_events` now iterates `self._event_subscriptions` for the event class directly instead of only currently-online peers — unsubscribe targets the pubsub service itself, not the peer, so it never depended on presence.
4. Added the two tests the plan claimed but didn't ship: `test_restart_resubscribes_to_already_online_peer` (step 5) and `test_module_opened_and_closed_events_still_fire` (step 6). 7 tests total, all passing against the live `tests/xmpp/docker-compose.yml` harness, plus the full existing XMPP suite (37/37) and unit suite (1452 passed, same pre-existing unrelated `test_pyobsd.py` failures as `develop`).

**Nits**
- ADR 0012: `proposed` → `accepted`.
- `specs/plans/index.md` entries brought in line with actual status.
- `logevent-double-delivery-fix-discussion.md` moved to `specs/plans/2026-08-16-logevent-double-delivery-fix-discussion.md`, with a status note pointing at the ADR since the shipped mechanism differs from what that discussion sketched.

thusser (commented):
Re-checked 68e71815: findings 1-4 and the ADR/index nits are all addressed correctly.

One remaining issue: the discussion doc was copied, not moved. Both `logevent-double-delivery-fix-discussion.md` (repo root) and `specs/plans/2026-08-16-logevent-double-delivery-fix-discussion.md` exist in the branch — the specs copy is the root copy plus a `Status:` note, and nothing references the root file anymore (ADR and plan index both point at the specs path). Delete the root copy to avoid a stale duplicate.

thusser:
Fixed in bb0bfc55 — removed the root-level copy, so only \`specs/plans/2026-08-16-logevent-double-delivery-fix-discussion.md\` remains.

For context on how that duplicate happened: the root file was committed to \`develop\` locally (commit 8df9d9c4) in an earlier part of this session, before I was asked to stop working in the local checkout and do the implementation in a separate clone instead. The clone inherited that commit, so the "move" only added the specs/plans copy without ever removing the original.

Note: your local \`develop\` checkout still has that same root-level file committed (also unpushed) from that earlier part of the session — I'm not touching your working copy per your earlier instruction, so flagging it here in case you want to clean it up separately.

thusser:
Non-blocking observation: a module subscribes to *every* online peer's node for *every* event type it has a handler for, so a camera with a `BadWeatherEvent` handler retry-subscribes to `admin:BadWeatherEvent` (and every other peer) forever, spamming `Still failing to subscribe ... after 30 attempts`. Since `_event_role` already advertises `send` vs `subscribe` in disco#info, gating subscriptions on peers that actually advertise `send` for that event would stop the noise.

thusser (commented):
Checks all pass and the four review findings + ADR/index nits are correctly addressed.

Two things before merge:

1. Drop the `graphify` commit (`75e4ba98`). It's 66k lines of unrelated `graphify-out/` churn (graph.json alone is 127k lines) bundled into an event-delivery PR. Split it into its own commit/PR so the real ~200-line change stays reviewable.

2. Stale PR body: says "4 new tests" but the file now has 7 test functions.

PR #762, opened 2026-08-16
Make pydantic config models reject unknown keys (extra="forbid")

thusser:
## Summary
- Sets `extra="forbid"` globally on `pyobs.utils.serialization.BaseModel`/`PolymorphicBaseModel`, so a misspelled or misplaced config key raises `ValidationError` at load instead of being silently dropped. Root cause: a task YAML put `guiding_config`/`acquisition_config` inside `instrument_configs` instead of one level up on `Configuration`; pydantic dropped both, the `Configuration`-level defaults applied instead, and the task failed `can_run` forever with no error pointing at the config.
- Declares the LCO portal models' previously-undeclared fields (`state`/`submitter` on `LcoSchedulableRequest`; `instrument_name`/`guide_camera_name`/`summary` on `LcoConfiguration`; eight fields on `LcoObservation`) instead of opting the family out with `extra="ignore"` — the portal is self-hosted, so a schema mismatch should fail loudly at upgrade time rather than being silently absorbed forever.
- `create_object`/`get_object` now route `comm`/`timezone`/`vfs`/`observer` through pydantic's validation `context` for pydantic models instead of passing them as constructor kwargs (which `extra="forbid"` would otherwise reject).
- `Task` is now a `PolymorphicBaseModel` so it pops its own `class` key.
- Fixed a previously-hidden bug surfaced along the way: `Merit.create()` left a stale `type` key in the config dict after deriving `class` from it.

Full design/decision history: `specs/plans/2026-08-15-pydantic-extra-validation.md`.

## Test plan
- [x] `pytest -m "not integration and not xmpp"`: 1460 passed, 25 skipped, 0 failed
- [x] `ruff check` clean on all changed files
- [x] `pyrefly check`: 0 errors
- [x] Added regression test (`tests/robotic/scripts/test_imaging.py`) reproducing the original misplaced-key bug and asserting it now raises

### DIFF
--- a/pyobs/object.py
+++ b/pyobs/object.py
@@ -186,7 +186,12 @@ def create_object(config: dict[str, Any], *args: Any, **kwargs: Any) -> Any:
     # pydantic models don't accept comm/timezone/vfs/observer as constructor kwargs (extra="forbid"
     # rejects them); route them through pydantic's context instead
     if issubclass(klass, BaseModel):
-        assert not args, "create_object() does not support positional args for pydantic models"
+        if args:
+            raise TypeError("create_object() does not support positional args for pydantic models")
+        overlap = set(cfg) & set(kwargs)
+        if overlap:
+            keys = ", ".join(sorted(overlap))
+            raise TypeError(f"create_object() got multiple values for keyword argument(s): {keys}")
         cfg = {**cfg, **kwargs}
         context = {p: cfg.pop(p, None) for p in ("comm", "timezone", "vfs", "observer")}
         return klass.model_validate(cfg, context=context, by_alias=True)
--- a/pyobs/object.py
+++ b/pyobs/object.py
@@ -186,9 +186,10 @@ def create_object(config: dict[str, Any], *args: Any, **kwargs: Any) -> Any:
     # pydantic models don't accept comm/timezone/vfs/observer as constructor kwargs (extra="forbid"
     # rejects them); route them through pydantic's context instead
     if issubclass(klass, BaseModel):
+        assert not args, "create_object() does not support positional args for pydantic models"
         cfg = {**cfg, **kwargs}
         context = {p: cfg.pop(p, None) for p in ("comm", "timezone", "vfs", "observer")}
-        return klass.model_validate(cfg, context=context)
+        return klass.model_validate(cfg, context=context, by_alias=True)
 
     # create object
     return klass(*args, **cfg, **kwargs)
--- a/pyobs/object.py
+++ b/pyobs/object.py
@@ -183,6 +183,13 @@ def create_object(config: dict[str, Any], *args: Any, **kwargs: Any) -> Any:
     cfg = copy.copy(config)
     del cfg["class"]
 
+    # pydantic models don't accept comm/timezone/vfs/observer as constructor kwargs (extra="forbid"
+    # rejects them); route them through pydantic's context instead
+    if issubclass(klass, BaseModel):
+        cfg = {**cfg, **kwargs}
+        context = {p: cfg.pop(p, None) for p in ("comm", "timezone", "vfs", "observer")}
+        return klass.model_validate(cfg, context=context)
+
     # create object
     return klass(*args, **cfg, **kwargs)
 
--- a/pyobs/robotic/scheduler/constraints/constraint.py
+++ b/pyobs/robotic/scheduler/constraints/constraint.py
@@ -45,6 +45,7 @@ class Constraint(PolymorphicBaseModel, metaclass=ABCMeta):
             except ValueError:
                 raise ValueError(f"Invalid constraint type: {config['type']}")
             config["class"] = f"pyobs.robotic.scheduler.constraints.{constraints[idx]}"
+            del config["type"]
         return obj.pyobs_model_validate(Constraint, config, by_alias=True)
 
     @staticmethod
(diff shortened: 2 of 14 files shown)


thusser:
The core change is sound and the tests back it up. One bug should be fixed in this PR, plus two things to close before merge.

**Blocking: `Constraint.create` has the same stale-`type` bug this PR fixes in `Merit.create`.** `pyobs/robotic/scheduler/constraints/constraint.py:47` sets `config["class"]` from `config["type"]` and never deletes `type`. `Constraint` is a `PolymorphicBaseModel`, so it now inherits `extra="forbid"`. Any scheduler config that lists global `constraints` as dicts with the `type` shorthand (`OnDemandScheduler(constraints=[{"type": "Airmass", ...}])`, `ondemandscheduler.py:59`) now raises `ValidationError` at load. No test catches it because every test passes `Constraint` *instances*, not `type` dicts. Add `del config["type"]` there, mirroring the `Merit.create` fix, plus a regression test.

**Verify before merge: required (no-default) fields on the LCO models.** `LcoObservation` gains `created`, `modified`, `ipp_value`, `name`, `observation_type`, `proposal`, `request_group_id`, `submitter` (all required), and `LcoSchedulableRequest` gains `state`, `submitter` (required). The plan itself notes the types are inferred from fixtures and should be confirmed against a live portal response. If any endpoint omits one (e.g. a not-yet-run observation without `ipp_value`), it hard-fails now. Confirm against real responses, or default the less-guaranteed fields.

**Rebase: the PR base is stale.** Recorded base is `04fd5187`, two commits behind `develop` (`7066350f`). Both doc commits are already on `develop`, so the UI shows 14 files / 3 commits when the true net diff is 9 files. Rebase onto `develop` so the review surface and merge match reality.

Minor, non-blocking:

- `create_object`'s new branch drops `*args` and calls `model_validate(cfg)` without `by_alias=True`. No current caller trips this, but it's a latent mismatch worth a comment or an explicit `by_alias=True`.
- `del config["type"]` in `Merit.create` sits inside `if "." not in config["type"]`, so the dotted-`type` branch still passes `type` through. That branch was already broken (the polymorphic model keys on `class`, not `type`), so it's dead code worth removing rather than leaving half-fixed.

Verified correct: `model_config` merges across inheritance (so `Task` and `LcoRequest` keep `extra="forbid"`), and the `comm`/`timezone`/`vfs`/`observer` context-routing in `create_object` is right, with no circular import.

thusser:
Re-checked the new head (`ec15dd75`). All review points are addressed:

- `Constraint.create` now deletes `type` after deriving `class` (`constraint.py:48`), plus a regression test (`test_constraint.py`).
- `Merit.create`'s dead dotted-`type` branch is gone; the `type` deletion is now unconditional, plus `test_merit.py`.
- `create_object`'s pydantic branch now passes `by_alias=True` and asserts against positional `*args`.
- Ran the affected tests against the branch: 59 passed (constraints/merits/imaging/LCO).

Two things left before merge:

1. **LCO required fields still unconfirmed.** `LcoObservation.created/modified/ipp_value/name/observation_type/proposal/request_group_id/submitter` and `LcoSchedulableRequest.state/submitter` are still required with no default. This is flagged in the plan as "needs the user" and can't be verified client-side. Confirm against the live self-hosted portal, or default the less-certain fields (e.g. `ipp_value`, `request_group_id`) so a not-yet-run observation doesn't hard-fail.

2. **The rebase landed on the stale tip, not current develop.** The plan says "rebased onto origin/develop's actual tip", but the branch base is `04fd5187`, two commits behind develop (`7066350f`). It's benign (`mergeable=true`, the diff is a clean 12 files with the doc commits correctly excluded), but the branch will show as 2 commits behind develop and the claim in the plan is inaccurate. Worth a `git rebase origin/develop` to land on `7066350f` before merge.

thusser:
Reviewed the full diff plus final state of every changed file; ran `tests/robotic` + `tests/utils` + `tests/test_object.py` (549 passed), `tests/robotic/storage/lco/` (57 passed), and `ruff check`/`pyrefly check` clean.

## Verdict: approve

Sound design, thorough plan, good regression coverage. Confirmed the core changes behave as claimed (`extra="forbid"`, declared LCO fields, context-routing in `create_object`, `Task` → `PolymorphicBaseModel`).

Findings, none blocking:

1. **`by_alias` dropped on polymorphic dispatch (pre-existing, dormant).** `retrieve_class_on_deserialization` calls `klass.model_validate(modified_value, context=info.context)` without `by_alias`, even though `Constraint.create`/`Merit.create`/`create_object` pass `by_alias=True`. Not triggered today: `create_object` deletes `class` before validating (so no dispatch), and no constraint/merit uses aliased fields. If one ever does, it'll fail validation. One-line fix in `serialization.py` if you want to close it now.

2. **`resolve_polymorphic_type_shorthand` mutates the caller's dict** (deletes `type`, adds `class`). Original also set `class` in place, but now it removes a key too. Low risk since configs are loaded fresh.

3. **LCO portal strictness is the real operational risk, by design.** Any undeclared field in a live portal response now raises at load. The self-hosted/fail-loud reclassification is reasonable and fields were checked against the portal source, but the plan still carries "confirm against a live portal response before finalizing" for the `LcoConfiguration` additions (inferred from fixtures). First portal upgrade that adds a field breaks pyobs at load.

4. **`Task.model_dump()` now emits `class` and `static_target` (field name, not `target` alias).** Not a regression (pydantic already dumps field names; `populate_by_name=True` makes both loadable; `class` is popped on deserialize). Just confirm no cross-repo backend expects a `class`-less task payload.

Nice side effect worth noting: `Task` created via `Object.get_object` now actually receives `comm`/`vfs`/`observer`/`timezone` through context, where `extra="ignore"` previously dropped them.

Issue #764, opened 2026-08-16
Fetching FITS header should never block anything


thusser:
2026-08-16 16:24:19 [WARNING] (fli230) fitsheader.py:108 Could not fetch FITS headers from admin: <RemoteTimeoutError> Call to get_fits_header_before on admin timed out..

Check whether this actually blocks anything.

thusser:
Fixed by #765, merged. Root cause: `FitsHeaderMixin.add_requested_fits_headers()` awaited every requested header future sequentially with no timeout, so a peer that never answers its IQ (in this case `admin`, a pyobs-gui on a laptop put to sleep without closing the client) stalled frame finalization for the full ~120s XMPP IQ timeout.

Fix bounds the whole collection to a single deadline (`fits_header_timeout`, default 15s) via `asyncio.wait()`, cancelling and skipping anything still pending.

Two additional bugs found during review and fixed in the same PR:
- `asyncio.wait()` crashed on an empty futures dict (any module with no comm or no peer implementing the FITS-header interfaces).
- `fits_header_timeout` wasn't reaching `BaseCamera`/`BaseVideo` (including `fli230`, the module that triggered this issue) due to explicit-kwarg-list constructors not forwarding it.

Follow-up filed as #767 for a separate, pre-existing gap (narrow `RemoteError` catch around the header fetch).

PR #765, opened 2026-08-16
Bound FITS-header fetch so a dead peer can't stall the frame

thusser:
## Summary

`FitsHeaderMixin.add_requested_fits_headers()` awaited every requested header future with no timeout. The futures are created before the exposure and awaited after readout, so a peer that never answers its IQ stalls frame finalization for the full XMPP IQ timeout (~120s — slixmpp's `Iq.send` default).

Trigger (issue #764): `admin` was a pyobs-gui on a laptop sent to sleep without closing the client. Its session lingered on ejabberd, and `fli230` spent ~120s per exposure fetching headers from it before giving up.

## Change

- Bound the whole collection to a single `asyncio.wait(..., timeout=...)` deadline; cancel + gather anything still pending, then skip it with a `timed out` warning (mirrors the existing `RemoteError` path).
- New `fits_header_timeout: float = 15.0` kwarg on `FitsHeaderMixin`, stored as `_fitsheadermixin_header_timeout`.

A `@timeout` decorator was considered and rejected: it works by having the *server* reply with a `method_timeout` IQ before running, and a dead peer never replies at all.

## Test plan

- `ruff check` / `black --check` clean; `pyrefly check` clean on `fitsheader.py`.
- `pytest tests/mixins/test_fitsheader.py`: 43 passed, including two new tests (dead peer times out and is cancelled; live+dead mix adds live headers and skips dead).

See `specs/plans/2026-08-16-fits-header-fetch-timeout.md` for the plan.

Note: this branch also contains commit `1e2b81d5` ("Mark core-tier test baseline plan implemented"), which the graphify automation committed to the branch on checkout — unrelated to this change.

### DIFF
--- a/pyobs/mixins/fitsheader.py
+++ b/pyobs/mixins/fitsheader.py
@@ -31,6 +31,7 @@ class FitsHeaderMixin:
         filenames: str = "/cache/pyobs-{DAY-OBS|date:}-{FRAMENUM|string:04d}.fits",
         frame_number: bool = True,
         night_obs: bool = True,
+        fits_header_timeout: float = 15.0,
         **kwargs: Any,
     ):
         """Initialise the mixin.
@@ -41,6 +42,9 @@ class FitsHeaderMixin:
             filename: Filename pattern for FITS images.
             frame_number: Whether to add frame number to FITS file header.
             night_obs: If True, DAY-OBS will contain the night of observation, not the calendar day.
+            fits_header_timeout: Maximum seconds to wait for a peer's FITS headers before skipping
+                them. A peer that never answers (e.g. a laptop put to sleep without closing its
+                client) would otherwise stall the frame for the full XMPP IQ timeout.
         """
         module = cast(Module, self)
 
@@ -48,6 +52,7 @@ class FitsHeaderMixin:
         self._fitsheadermixin_fits_namespaces = fits_namespaces
         self._fitsheadermixin_filename_pattern = filenames
         self._fitsheadermixin_fits_headers = fits_headers if fits_headers is not None else {}
+        self._fitsheadermixin_header_timeout = fits_header_timeout
         if "OBSERVER" not in self._fitsheadermixin_fits_headers:
             self._fitsheadermixin_fits_headers["OBSERVER"] = ["pyobs", "Name of observer"]
         self._fitsheadermixin_night_obs = night_obs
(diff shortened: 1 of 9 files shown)


thusser:
Found and fixed two bugs while reviewing this (commit caa56218):

1. **Crash on empty `futures`.** `asyncio.wait(futures.values(), timeout=...)` raises `ValueError: Set of Tasks/Futures is empty.` whenever `futures` is `{}` — which `request_fits_headers()` returns for any module with no comm, or with comm but no peer implementing `IFitsHeaderBefore`/`IFitsHeaderAfter` (dev/test setups, single-module deployments). Previously a harmless no-op loop, now crashes every exposure on that path. None of the existing tests (old or new) exercised an empty dict, and the one test that runs a real `grab_data()` (`test_aborted_exposure_raises_aborted_error`) aborts before reaching header assembly, so this shipped uncaught. Fixed by skipping the `asyncio.wait` call when `futures` is empty.

2. **`fits_header_timeout` never reaches `BaseCamera`/`BaseVideo`.** Both call `ImageFitsHeaderMixin.__init__` with an explicit keyword list rather than forwarding `**kwargs` (unlike `basespectrograph.py`, which does), so a configured `fits_header_timeout` silently fell into `Module`/`Object`'s catch-all kwargs and was dropped — defeating the PR's stated purpose for `fli230` (a `BaseCamera`), the exact module in the motivating incident. Fixed by adding an explicit `fits_header_timeout` param to both and threading it through.

Added regression tests for both, plus a `fits_header_timeout`-reaches-mixin test for `BaseSpectrograph` (that path was already correct via `**kwargs`, just untested). Full `tests/mixins/` + `tests/modules/camera/` suite (134 tests) passes; ruff/black/pyrefly clean.

thusser (commented):
Verified on head caa56218: pytest (fitsheader + basecamera/basespectrograph/basevideo) 87 passed, ruff/black/pyrefly clean.

Fix is correct and a real improvement (concurrent await under one deadline vs the old sequential awaits). Two non-blocking notes:

1. `future.result()` only guards `RemoteError`. A peer that errors with something else (or a task cancelled externally) still propagates. Same as the old `await`, so no regression, but worth a follow-up if peers ever fail with non-RemoteError.

2. `1e2b81d5` (test-baseline plan 'implemented') rides along here and touches specs/. Drop it if you want a clean single-purpose PR; otherwise harmless doc churn.

LGTM.

PR #773, opened 2026-08-17
Fix comm_cfg anchor-holder key leaking into config dicts

thusser:
## Summary
- `pre_process_yaml`'s whole-file `{include file}` spliced the entire included file in, including keys whose sole purpose is holding a YAML anchor for `<<: *anchor` use elsewhere -- e.g. `comm.shared.yaml`'s `comm_cfg: &comm`, used across every monti/monet/iagvt/polaris config. That leaked `comm_cfg` reached `Object.__init__`'s `**kwargs` and was silently dropped, indistinguishable from a real config typo.
- Whole-file includes now drop any key `reload_anchors()` identifies as an anchor holder for that file before splicing; keyed includes of the same key (`{include file key}`) are untouched.
- Handles the resulting edge case where the entire include is anchor-only keys (`comm.shared.yaml`'s only top-level key), by dropping the splice placeholder instead of emitting an empty `{}` mapping, which broke YAML parsing when followed by block-style content.
- Verified against a real `pyobs-monet` config (`config/central/imagedb.yaml`): `comm_cfg` no longer leaks, `comm: <<: *comm` still resolves correctly.

Implements the `comm_cfg` portion of `specs/plans/2026-08-09-object-kwarg-validation.md`. The `environment`/`database` wrapper-key question and `Object.__init__`'s warn/raise enforcement are intentionally left open -- see that plan's Decision section for why.

## Test plan
- [x] `tests/utils/test_config.py`: 3 new tests -- alias resolution (previously uncovered), keyed-include-of-anchor-holder is kept, whole-file-include no longer leaks the anchor holder
- [x] Full suite: `pytest -m "not integration and not xmpp"` -- 1488 passed, 25 skipped, 0 failed
- [x] `ruff check` / `black --check` / `pyrefly check` clean on changed files
- [x] Manual check against `pyobs-monet/config/central/imagedb.yaml`

### DIFF
--- a/pyobs/utils/config.py
+++ b/pyobs/utils/config.py
@@ -25,18 +25,36 @@ def pre_process_yaml(config: str) -> str:
     pattern = r"((\s*)?(-\s*)?\{include (\S*)( \S*)?\})"
     matches = re.findall(pattern, content)
     for match, indent, tick, filename, key in matches:
+        matches_anchor = reload_anchors(path + "/" + filename)
         with StringIO(pre_process_yaml(path + "/" + filename)) as f:
             include_full = yaml.safe_load(f)
             include_dict = include_parts(include_full, key)
-        include = yaml.dump(include_dict, default_flow_style=False, indent=2)
 
-        # ensure indentation level to be conserved
-        if tick != "":
-            include = tick + include
-        if indent != "":
-            indent_newline = indent + " " * len(tick)
-            include = indent + include.replace("\n", indent_newline)
-        matches_anchor = reload_anchors(path + "/" + filename)
+        # a whole-file include (no key selector) would otherwise leak any key whose sole purpose
+        # is holding an anchor for `<<: *anchor` use elsewhere (e.g. comm.shared.yaml's
+        # `comm_cfg: &comm`) as an unconsumed top-level key in the final config. Drop those here;
+        # `<<: *anchor` below still resolves correctly, since replace_aliases reads the anchor's
+        # value from the original included file, not from this (possibly trimmed) copy.
+        if not key.strip() and isinstance(include_dict, dict):
+            anchor_keywords = {keyword for keyword, _anchor in matches_anchor}
+            include_dict = {k: v for k, v in include_dict.items() if k not in anchor_keywords}
+
+        if isinstance(include_dict, dict) and not include_dict:
+            # everything in this whole-file include was anchor-holder keys (e.g. a
+            # comm.shared.yaml whose only top-level key is `comm_cfg: &comm`) -- nothing left to
+            # splice in. yaml.dump({}) is "{}\n", a flow-style node that isn't a valid sibling to
+            # the block-style mapping the rest of the file continues as, so drop the placeholder
+            # entirely instead of inserting an empty mapping.
+            include = ""
+        else:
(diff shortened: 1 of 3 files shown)


thusser:
Approach is sound and it fixes the `comm.shared.yaml` case, but two edge cases need attention before merge.

**1. Empty-dict splice isn't scoped to whole-file includes** (`pyobs/utils/config.py:42`)

The `isinstance(include_dict, dict) and not include_dict` branch fires for *keyed* includes too, not just the anchor-trimmed whole-file case. A keyed include that legitimately selects an empty mapping now degrades:

```yaml
# inc.yaml
somekey: {}
# main.yaml
outer:
  {include inc.yaml somekey}
```

Before: `outer: {}`. After: `outer: null` (the `{}` placeholder is dropped). The `not key.strip()` guard on line 38 only protects the anchor-dropping block, not this one. Gate line 42 with the same whole-file condition (or track whether the dict was trimmed) so keyed empty-dict includes keep emitting `{}`.

**2. Anchor-holder detection isn't top-level-scoped** (`pyobs/utils/config.py:39`)

`reload_anchors`'s regex `(\S*): &(\S*)` also captures *nested* anchors, and `keyword` is just the last non-space token before `: &`, not a top-level path. So a nested key that shares a name with a top-level key causes the top-level key to be silently dropped from a whole-file include:

```yaml
# inc.yaml
type: Foo
camera:
  type: &cam
    model: X
```

After a whole-file include, top-level `type: Foo` is gone (only `camera` survives). Restrict the drop to keys that are actual top-level anchor holders, not name-matched against every anchor anywhere in the file.

Both are silent-data-loss shapes, i.e. exactly the class of bug this PR is meant to kill. They're unlikely in today's fleet configs (only `comm.shared.yaml` is whole-file included, and its lone anchor is top-level), but worth closing with a regression test each.

Otherwise good: `tests/utils/test_config.py` passes (19 passed), the `xfail` removal is correct, the new `replace_aliases` coverage fills a real gap, and the plan doc update accurately leaves `environment`/`database` and the `Object.__init__` enforcement open.

thusser:
Both fixed in 81244679.

**1. Empty-splice drop now scoped to whole-file includes only.** Gated that branch on the same `not key.strip()` condition as the anchor-drop itself, so a keyed include of a legitimately-empty mapping (`{include file key}` where `key`'s value is `{}`) still emits `{}` instead of being silently dropped to `null`.

**2. Anchor-holder detection now top-level-scoped.** Added `top_level_anchor_keywords()`, a line-anchored (`^`, no leading whitespace) regex restricted to unindented keys, used only for the whole-file drop decision. `reload_anchors()` itself is untouched -- `replace_aliases()` still needs it to resolve anchors at any nesting depth, so changing its general behavior wasn't an option.

Reproduced both bugs from your examples before fixing, to confirm they were real. Added a regression test for each (`test_pre_process_yaml_keyed_empty_dict_include_is_not_dropped`, `test_pre_process_yaml_whole_file_include_keeps_top_level_key_matching_nested_anchor_name`). Full suite: 1490 passed, 25 skipped. Re-ran the 803-file scan across pyobs-monet/pyobs-iagvt/pyobs-iag50 -- same result as before (2 pre-existing, unrelated errors; no `comm_cfg` leaks in any consuming config).

Issue #19, opened 2020-10-27
gui crashes when trying to navigate to old commands, but none were given before


masurt:
when no cammands were issued yet in the shell widgets and you try to go back to the last command by pressing arrow up, the gui crashes.
```
Traceback (most recent call last):
  File "/home/tilman/autoguiding/pyobs-gui/pyobs_gui/qt/commandinput.py", line 29, in keyPressEvent
    self.setText(self._history[self._history_index])
IndexError: list index out of range
Aborted (core dumped)
```

thusser:
fixed

PR #776, opened 2026-08-18
Make pyobs-core's mixin composition cooperative (step 1/10, cooperative-mixin-init plan)

thusser:
## Summary

First step of `specs/plans/2026-08-18-cooperative-mixin-init.md`: makes `pyobs-core`'s own mixin composition genuinely cooperative, so `Object.__init__` can eventually raise on unrecognized kwargs (attempted directly in `specs/plans/2026-08-09-object-kwarg-validation.md`, reverted -- broke 59 tests because several classes fan the same `**kwargs` out to multiple sibling mixins independently instead of a `super()` chain).

- `Object.__init__` now ends with `super().__init__(**kwargs)` instead of silently absorbing leftovers.
- Converted the 14 `pyobs-core` classes (+ 3 test doubles) using the explicit fan-out pattern to a single cooperative `super().__init__()` call.
- Reordered bases in `BaseRoof`, `BaseTelescope`, `DummyMode` to put `Module` first -- every order dependency found (mixins calling `add_background_task`/`add_child_object`/`self.comm`) needs `Object`/`Module` to run first, none need the opposite.
- Made `FitsHeaderMixin`'s cache-path attribute a lazy property instead of computing it eagerly at `__init__` time, since it can now run before `Module` finishes setting up `self._device_name`.
- Made `PipelineMixin`'s `archive` default-injection conditional on the target step actually declaring the parameter (signature-checked), instead of injecting unconditionally and relying on silent-drop -- the exact thing this whole effort removes.

## Real bugs found and fixed along the way

A full fleet-config construction check (not just unit tests -- actually building real objects from 799 config files across `pyobs-monet`/`pyobs-iagvt`/`pyobs-iag50`/`pyobs-polaris`) surfaced two previously-hidden, silently-dropped kwargs in `pyobs-core` itself:

- `Scheduler` passed `auto_update=False` to `ObservationArchive`/`LcoObservationArchive`, which have never declared that parameter. Removed.
- `Scheduler` passed `observation_archive` to every `TaskScheduler` implementation, but only `OnDemandScheduler` declares it -- `AstroplanScheduler` (the one actually configured in the fleet) always dropped it. Now only injected when the configured class actually declares it.
- `BasePointing`'s docstring documented `log_file`/`log_absolute` as if they were its own params; never implemented there (only on the nested `apply:` object, and on `Acquisition` specifically). Removed the false documentation.

Companion fleet-config fixes (moving misplaced `log_file`/`max_offset` into the `apply:` block where they're actually consumed, removing one true duplicate) are in separate PRs against `pyobs-monet` and `pyobs-iag50`.

## Test plan
- [x] Full suite: `pytest -m "not integration and not xmpp" --extra full` -- 1490 passed, 25 skipped, 0 failed
- [x] `ruff check` / `black --check` / `pyrefly check` clean
- [x] Manually verified the double-init bug this could have hidden is actually fixed (counted `FitsHeaderMixin.__init__` calls before/after: 1, not 2)
- [x] Constructed all 799 real fleet config files across 4 repos; every previously-constructible config still constructs, the two real bugs above were the only new failures (both fixed)

### DIFF
--- a/pyobs/mixins/camerasettings.py
+++ b/pyobs/mixins/camerasettings.py
@@ -35,6 +35,8 @@ class CameraSettingsMixin:
         self.__camerasettings_filter = filter_name
         self.__camerasettings_binning = binning
 
+        super().__init__(**kwargs)
+
     async def _do_camera_settings(self, camera: Module | IData | IFilters | IBinning | IWindow) -> None:
         """Do camera settings for given camera."""
 
--- a/pyobs/mixins/fitsheader.py
+++ b/pyobs/mixins/fitsheader.py
@@ -46,8 +46,6 @@ class FitsHeaderMixin:
                 them. A peer that never answers (e.g. a laptop put to sleep without closing its
                 client) would otherwise stall the frame for the full XMPP IQ timeout.
         """
-        module = cast(Module, self)
-
         # store
         self._fitsheadermixin_fits_namespaces = fits_namespaces
         self._fitsheadermixin_filename_pattern = filenames
@@ -58,9 +56,18 @@ class FitsHeaderMixin:
         self._fitsheadermixin_night_obs = night_obs
 
         # night exposure number
-        self._fitsheadermixin_cache = f"/pyobs/modules/{module.name}/cache.yaml"
         self._fitsheadermixin_enable_frame_number = frame_number
         self._fitsheadermixin_frame_number = 0
+
+        super().__init__(**kwargs)
+
+    @property
+    def _fitsheadermixin_cache(self) -> str:
+        # computed lazily (not cached at __init__ time) since it depends on module.name, which in
+        # turn depends on comm having been set up by Object.__init__ -- in a cooperative super()
+        # chain, this mixin's __init__ can run before Module has finished its own setup
+        module = cast(Module, self)
+        return f"/pyobs/modules/{module.name}/cache.yaml"
 
     async def request_fits_headers(self, before: bool = True) -> dict[str, Task[Any]]:
         """Request FITS headers from other modules.
@@ -309,7 +316,7 @@ class ImageFitsHeaderMixin(FitsHeaderMixin):
             rotation: Rotation east of north.
             filename: Filename pattern for FITS images.
         """
-        FitsHeaderMixin.__init__(self, **kwargs)
+        super().__init__(**kwargs)
 
         # store
         self._fitsheadermixin_centre = centre
--- a/pyobs/mixins/fitsnamespace.py
+++ b/pyobs/mixins/fitsnamespace.py
@@ -15,6 +15,7 @@ class FitsNamespaceMixin:
 
     def __init__(self, fits_namespaces: dict[str, list[str]] | None = None, **kwargs: Any):
         self.__namespaces = {} if fits_namespaces is None else fits_namespaces
+        super().__init__(**kwargs)
 
     def _filter_fits_namespace(
         self, hdr: dict[str, FitsHeaderEntry], sender: str, namespaces: list[str] | None = None, **kwargs: Any
--- a/pyobs/mixins/follow.py
+++ b/pyobs/mixins/follow.py
@@ -125,6 +125,8 @@ class FollowMixin:
         if not isinstance(self, self.__follow_mode):
             raise ValueError(f"This module is not of given mode {mode}.")
 
+        super().__init__(*args, **kwargs)
+
     @property
     def is_following(self) -> bool:
         """Returns True, if we're following another device."""
--- a/pyobs/mixins/motionstatus.py
+++ b/pyobs/mixins/motionstatus.py
@@ -26,6 +26,8 @@ class MotionStatusMixin:
         self.__motion_status = MotionStatus.UNKNOWN
         self.__motion_status_single = {i: MotionStatus.UNKNOWN for i in self.__motion_status_interfaces}
 
+        super().__init__(**kwargs)
+
     async def open(self) -> None:
         # subscribe to events
         if isinstance(self, Module) and self._comm:
--- a/pyobs/mixins/pipeline.py
+++ b/pyobs/mixins/pipeline.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
+import inspect
 import logging
 from typing import TYPE_CHECKING, Any
 
 from pyobs.images import Image, ImageProcessor
-from pyobs.object import Object
+from pyobs.object import Object, get_class_from_string
 from pyobs.utils.exceptions import ImageError
 
 if TYPE_CHECKING:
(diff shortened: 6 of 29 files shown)


thusser:
## Review: verified mechanically sound, but one real production regression found

I checked out the head (`2b95b6f0`) and verified the conversion by running it: MROs put `Object` directly after `Module` and before every mixin, so `Object.__init__`'s `super()` correctly reaches them; all converted classes construct with realistic kwargs (`DummyRoof`, `DummyRaDecTelescope`, `DummyCamera`, `PipelineCamera`, `AutoFocusSeries`, `Acquisition`, `AutoGuiding`, both `Pipeline`s); the double-init is genuinely fixed (monkey-patched `ImageFitsHeaderMixin.__init__` call-count = **1**); the lazy `_fitsheadermixin_cache` property works and is never assigned; both signature helpers behave correctly; full suite passes locally (the 7 failures are environmental — read-only `/opt/pyobs/storage` — and fail identically on `develop`).

### Finding 1 (regression): removing `auto_update=False` re-enables a background poller in two live fleets

`pyobs/modules/robotic/scheduler.py` drops `auto_update=False` from the schedule-archive construction with the justification that it was *"never declared by ObservationArchive/LcoObservationArchive — silently dropped, no effect, for this class's entire history."* That claim is wrong for `BackendObservationArchive`:

- `pyobs/robotic/storage/backend/observationarchive.py:26` **does declare** `auto_update: bool = True` and gates a `_check_for_changes` background task on it — a **5-second polling loop** against the backend API.
- Two **live** fleet configs use exactly that class as `schedule:`: `pyobs-monet/config/south/monet/scheduler.yaml` and `pyobs-iagvt/config/iagvtsrv/scheduler.yaml`.
- The kwarg was **deliberately** introduced in `49b8b035` ("rely on on_tasks_changed for task changes", 2026-05-15), which also removed the startup `_update_schedule()` task — i.e. the scheduler moved to event-driven updates and explicitly disabled the archive's own poller.

Runtime demonstration with the fleet config shape:

| | background tasks on the schedule archive |
|---|---|
| `develop` | `[]` — poller disabled |
| this branch | `[BackgroundTask]` — `_check_for_changes` polling every 5 s |

The fleet construction check couldn't catch this — construction succeeds either way; it's a runtime behavior change. Ironically it's exactly the "silently-dropped kwarg that was actually load-bearing" class of bug this PR set out to eliminate.

**Suggested fix** (fits the PR's own pattern): inject `auto_update=False` only when the configured class declares it, same shape as `_scheduler_accepts_observation_archive`. It can't be passed unconditionally anymore — `LcoObservationArchive` doesn't declare it, and the leftover would now hit `object.__init__()` → `TypeError`.

### Finding 2: unrecognized-kwarg enforcement is already live, with a confusing message

`Object.__init__` forwards leftovers to `super().__init__(**kwargs)` → `object.__init__`, so **any** unrecognized kwarg on **any** `Object` subclass now raises `TypeError: object.__init__() takes exactly one argument (the instance to initialize)` — confirmed for `Module(foo=1)`, `DummyCamera(foo=1)`, plain `Object(...)` (all silently absorbed on `develop`). The PR frames raising as a future step ("can eventually raise"), but it's already happening — on every plugin class too, with a message that gives no hint which kwarg is wrong. The plan documents the cross-repo blast radius (pyobs-fli #84 / pyobs-gemini #19 fixed; 6 repos latent). Worth an explicit decision on whether step 1 should ship this already, and ideally fast-tracking step 2's proper `raise TypeError` that names the offending kwarg.

### Finding 3 (minor): `_scheduler_accepts_observation_archive` ignores string configs

It returns `False` unless `scheduler` is a dict, so a scheduler configured as a bare class-path string silently loses its `observation_archive` — a behavior change vs. the old unconditional injection. No live config uses string form (all dict), and the param defaults to `None` on `OnDemandScheduler`, so nothing breaks today; `PipelineMixin._accepts_archive` already handles class-name strings and the scheduler helper should too.

### Suggested follow-up

Add a unit test pinning the `Scheduler` kwarg-injection matrix: `BackendObservationArchive` schedule → gets `auto_update=False`; `LcoObservationArchive` → doesn't; `OnDemandScheduler` dict → gets `observation_archive`; `AstroplanScheduler` → doesn't. The two real bugs this PR found were caught only by a manual fleet-construction check — these code paths currently have zero test coverage, which is how Finding 1 slipped through.

Otherwise: excellent work. The conversion is careful and well-documented, and the plan's blast-radius analysis is honest and thorough.

thusser:
Fixed Finding 1 in \`422bfcd6\`.

- Generalized \`_scheduler_accepts_observation_archive\` into \`_class_accepts_param(config, param_name)\`, now used for both the \`observation_archive\` and \`auto_update\` injections. Also extended it to accept a bare class object (not just a config dict), matching what \`add_child_object\` itself accepts (partially addresses Finding 3 for this code path -- Scheduler's own type hints never allow a string class-path here, so that specific case doesn't apply).
- \`auto_update=False\` is now injected only when the configured \`schedule\` class actually declares it (true for \`BackendObservationArchive\`, false for \`LcoObservationArchive\`), instead of being dropped unconditionally.
- Added a kwarg-injection matrix test: \`_class_accepts_param\` against all four real classes (dict-config and bare-class forms), plus an end-to-end construction test confirming \`BackendObservationArchive\`'s \`_check_for_changes\` poller stays off when constructed through \`Scheduler\`.

Finding 2 (the already-live \`object.__init__()\` enforcement) and Finding 3's broader point (the pattern's general string-handling, elsewhere) are left as-is for now -- happy to open follow-up issues for those rather than block this PR further, if that works.

thusser:
Verified the fix on `422bfcd6` — Finding 1 is properly closed.

- Re-ran the runtime check against the fleet config shape: through `Scheduler`, `BackendObservationArchive`'s schedule archive now has **0 background tasks** (poller off, matching `develop`), and `observation_archive` is still injected for a dict-configured `OnDemandScheduler`.
- All 42 tests in `tests/modules/robotic/test_scheduler.py` pass locally, including the new matrix (`BackendObservationArchive`→`auto_update` True, `LcoObservationArchive`→False, `OnDemandScheduler`→`observation_archive` True, `AstroplanScheduler`→False, in both dict-config and bare-class forms).
- One edge I double-checked: `_class_accepts_param` can now return True for an already-built instance, but `get_object` ignores kwargs for instances (returns the object as-is), so injecting `auto_update=False` there is a safe no-op — no crash path.

Agree with leaving Finding 2 and Finding 3's broader point for follow-up issues rather than blocking this PR. When you open them, the two concrete asks worth capturing: (a) the already-live enforcement at `Object.__init__` surfaces as the raw `TypeError: object.__init__() takes exactly one argument (the instance to initialize)` — step 2's proper `raise TypeError` naming the offending kwarg would fix both the message and the "step 1 already enforces" surprise; (b) the string-config gap is general (any `_class_accepts_param`-style caller that only inspects dicts), even if `Scheduler`'s own type hints never allow a bare class-path string.

Thanks for the fast turn on this — the regression test makes it stick.

PR #782, opened 2026-08-19
Give leftover-kwargs TypeError in Object.__init__ a diagnosable message

thusser:
## Summary

Closes the last non-operational item from `specs/plans/2026-08-18-cooperative-mixin-init.md`'s
checklist (`Implement if kwargs: raise TypeError(...) in Object.__init__`).

`Object.__init__`'s cooperative `super().__init__(**kwargs)` call, if nothing further down
the chain claims a kwarg, eventually reaches the real `object.__init__()`, which raises
`"object.__init__() takes exactly one argument (the instance to initialize)"` -- a message
that names neither the offending class nor which kwarg leaked. The cooperative-mixin-init
rollout this closes out repeatedly needed multi-frame traceback tracing to diagnose exactly
that.

**Deliberately not a blind `if kwargs: raise` at the top of `Object.__init__`**: `BaseRoof`'s
real MRO (`Module -> Object -> PrivateAttrMixin -> ... -> WeatherAwareMixin ->
MotionStatusMixin -> ... -> object`, confirmed empirically) places mixins *after* `Object`
that still need to consume their own kwargs cooperatively -- `Object` is not always the last
stop before `object.__init__()`. A naive check there would break every `BaseRoof`-derived
class's ability to thread `motion_status_interfaces` etc. through the chain.

Instead, this wraps the `super().__init__(**kwargs)` call and re-raises with a clear message
(naming the class and the kwargs `Object` itself never recognized) only when the underlying
`TypeError` is `object.__init__()`'s own generic one, chaining the original via `from e` so
nothing is lost.

## Test plan

- [x] New test confirming the error names the class and kwarg
- [x] New test confirming the original `TypeError` is chained (`__cause__`)
- [x] New regression test confirming a mixin listed *after* `Object` in a subclass's MRO can
      still consume its own kwargs cooperatively (guards against the naive-check regression)
- [x] Full suite: 1504 passed, 0 failed
- [x] `ruff check`, `black --check`, `pyrefly check` all clean

### DIFF
--- a/pyobs/object.py
+++ b/pyobs/object.py
@@ -357,8 +357,19 @@ class Object(PrivateAttrMixin):
 
         # forward anything left to the next class in the MRO -- lets mixins listed after this
         # one in a subclass's bases (e.g. WeatherAwareMixin, PipelineMixin) claim their own
-        # kwargs cooperatively, instead of Object silently absorbing them
-        super().__init__(**kwargs)
+        # kwargs cooperatively, instead of Object silently absorbing them. If nothing further
+        # down the chain claims them either, this eventually raises inside the real
+        # object.__init__(), whose error message names neither the class nor the offending
+        # kwargs -- wrap it so a leftover-kwarg mistake is actually diagnosable.
+        try:
+            super().__init__(**kwargs)
+        except TypeError as e:
+            if kwargs and "object.__init__() takes exactly one argument" in str(e):
+                raise TypeError(
+                    f"{type(self).__name__}() got unexpected keyword argument(s) "
+                    f"{sorted(kwargs)} that no class in its __init__ chain consumed."
+                ) from e
+            raise
 
     def add_background_task(
         self, func: Callable[..., Coroutine[Any, Any, None]], restart: bool = True, autostart: bool = True
--- a/pyobs/object.py
+++ b/pyobs/object.py
@@ -365,9 +365,23 @@ class Object(PrivateAttrMixin):
             super().__init__(**kwargs)
         except TypeError as e:
             if kwargs and "object.__init__() takes exactly one argument" in str(e):
+                # Object's own kwargs is only the leftover *before* consumption by whatever mixins
+                # come after it in the MRO (e.g. MotionStatusMixin claiming motion_status_interfaces)
+                # -- reporting it as-is would flag kwargs that were, in fact, legitimately consumed
+                # downstream. The precise set is the `kwargs` local of the deepest frame in this
+                # exception's traceback: that's the cooperative __init__ call that actually forwarded
+                # to the real object.__init__() and got rejected, so its own unconsumed kwargs is
+                # exactly what's left over.
+                leftover = kwargs
+                tb = e.__traceback__
+                while tb is not None:
+                    frame_kwargs = tb.tb_frame.f_locals.get("kwargs")
+                    if isinstance(frame_kwargs, dict):
+                        leftover = frame_kwargs
+                    tb = tb.tb_next
                 raise TypeError(
                     f"{type(self).__name__}() got unexpected keyword argument(s) "
-                    f"{sorted(kwargs)} that no class in its __init__ chain consumed."
+                    f"{sorted(leftover)} that no class in its __init__ chain consumed."
                 ) from e
             raise
 
--- a/specs/plans/2026-08-09-object-kwarg-validation.md
+++ b/specs/plans/2026-08-09-object-kwarg-validation.md
@@ -1,14 +1,22 @@
 # Plan: Surface unrecognized kwargs in `Object.__init__` instead of silently discarding them
 
-Status: `comm_cfg` anchor leak fixed at the source, merged 2026-08-17 (PR #773, squash-merged as
-`a5646fb8`); `environment`/`database` confirmed gone 2026-08-18 (no longer a blocker); two fleet
-cleanup passes (2026-08-18) fixed every confirmed dead/misplaced/typo'd kwarg found across
-`pyobs-monet`/`pyobs-iagvt`/`pyobs-iag50`/`pyobs-polaris`, including chasing down the ~45 classes
-that couldn't be import-checked in the first pass. **A first attempt at implementing the raise in
-`Object.__init__` (2026-08-18) found a real architectural blocker — see "Raise attempt" below — and
-was reverted.** The prerequisite fix now has its own plan:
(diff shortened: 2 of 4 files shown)


thusser:
Reviewed against the merged `develop` (parent of this commit == current `develop` head; the whole rollout has landed — #776 merged 04:17Z, all six downstream PRs/MRs merged, `v2.0.0.dev82` on PyPI — so this capstone lands at exactly the right time).

**Verdict: approve.** Well-scoped, well-tested, and the design rationale is exactly right.

Verified:
- New tests pass (9/9 in `test_object.py`, plus `tests/modules/roof`); `ruff` and `black --check` clean.
- The wrap correctly leaves genuine downstream `TypeError`s untouched (verified: a missing-required-arg error in a later `__init__` is re-raised as-is, not rewritten) and chains the original via `__cause__`.
- The mixin-after-Object regression guard mirrors the real `BaseRoof` MRO (`Module -> Object -> PrivateAttrMixin -> ... -> WeatherAwareMixin -> MotionStatusMixin -> ... -> object`), which I'd previously confirmed empirically — so the "not a blind `if kwargs: raise`" reasoning is sound, and a naive check at `Object` would indeed have broken every `BaseRoof`-derived chain.

One finding (non-blocking) — **the reported kwarg set can include kwargs a downstream class legitimately consumed**:

`Object.__init__` reports its *own* pre-consumption leftover, but the kwargs that actually reach `object.__init__()` are a strict subset (downstream `__init__`s bind theirs as named params). Repro:

```python
class MyRoof(Module, MotionStatusMixin):
    pass

MyRoof(motion_status_interfaces=["IRoof"], bogus_kwarg="x")
# → "MyRoof() got unexpected keyword argument(s) ['bogus_kwarg', 'motion_status_interfaces'] that no class in its __init__ chain consumed."
```

`motion_status_interfaces` *was* consumed — by `MotionStatusMixin`. In the fleet's typical failure mode (a typo'd config kwarg on a class whose chain hardcodes `motion_status_interfaces`/`wait_for_*`), the message will flag kwargs the user never even passed, and the "that no class in its __init__ chain consumed" phrasing is then factually wrong about most of the list. Two ways to tighten it:
1. Derive the precise set from the chained exception's traceback — the deepest cooperative-`__init__` frame's `kwargs` local is exactly the final leftover — or
2. At minimum reword to something that doesn't over-claim, e.g. "`Object.__init__` forwarded these down its cooperative chain and `object.__init__()` rejected the leftovers: [...]".

Nit: the plan doc's checklist item this PR closes (`specs/plans/2026-08-18-cooperative-mixin-init.md` line ~444, "Implement `if kwargs: raise TypeError(...)` ... Not yet implemented") is implemented here in the wrapped form, but the doc isn't updated — worth checking it off (and the item at line ~456 re: closing `object-kwarg-validation.md`'s Status is now unblocked by this PR).

thusser:
Thanks — the plan-doc closure commit (`53125e17`) addresses my nit exactly: both `2026-08-09-object-kwarg-validation.md` and `2026-08-18-cooperative-mixin-init.md` now have accurate `Status: implemented, closed` lines, the enforcement checklist items are checked off with the wrap-vs-blind-check distinction recorded, and the one remaining open item (restarting live fleet modules) is correctly kept out of code/doc scope.

That leaves only the one non-blocking suggestion from my previous comment, restated since it's now the only open thread: the error message reports `Object.__init__`'s *pre-consumption* leftover, so when a class's chain legitimately claims some of it downstream (e.g. `motion_status_interfaces` consumed by `MotionStatusMixin`) and a different kwarg leaks, the message flags the legitimately-consumed ones too ("…that no class in its __init__ chain consumed" is then inaccurate). Repro: `MyRoof(motion_status_interfaces=["IRoof"], bogus_kwarg="x")` → lists both `['bogus_kwarg', 'motion_status_interfaces']`.

If you want to tighten it before merge: the exact set that reached `object.__init__()` is recoverable from the chained exception's traceback — the deepest cooperative-`__init__` frame's `kwargs` local is precisely the final leftover — so the message could name only the real offenders. Alternatively, rewording to "`Object.__init__` forwarded these down its cooperative chain and `object.__init__()` rejected the leftovers" removes the over-claim without touching the traceback. Either way the fix is strictly better than today's "object.__init__() takes exactly one argument", so this is not a blocker from my side — approve as-is or with the tightening.

thusser:
Tightened it via traceback introspection rather than just rewording. In the except block, `e.__traceback__`'s deepest frame is the cooperative `__init__` call that actually forwarded to the real `object.__init__()` and got rejected -- its own `kwargs` local at that point is exactly the precise leftover set, since any class between `Object` and there (like `MotionStatusMixin`) has already bound its own named params out of it.

Walk `e.__traceback__` via `tb_next` to the last frame with a dict-typed `kwargs` local, use that instead of `Object`'s own (pre-consumption) `kwargs`.

Verified against your exact repro:

```python
class MyRoof(Module, MotionStatusMixin):
    pass

MyRoof(motion_status_interfaces=["IRoof"], bogus_kwarg="x")
# now: "MyRoof() got unexpected keyword argument(s) ['bogus_kwarg'] that no class in its __init__ chain consumed."
```

Added a dedicated test for this (downstream-consumed kwarg alongside a real typo, asserting the message contains the typo but not the legitimately-consumed one). Full suite: 1505 passed, ruff/black/pyrefly clean.

thusser:
Verified the fix — it does exactly what it claims. On the updated head (`9177b9f1`):

- Your exact repro shape now reports only the real offender: `MyRoof(motion_status_interfaces=["IRoof"], bogus_kwarg="x")` → "got unexpected keyword argument(s) `['bogus_kwarg']`", with `motion_status_interfaces` correctly absent.
- The new `test_unrecognized_kwarg_message_excludes_downstream_consumed_kwargs` passes; full `test_object.py` 10/10; `tests/modules/roof` + `tests/mixins` clean (the one failure there is the pre-existing `/opt/pyobs/storage` read-only sandbox issue, unrelated); `ruff`/`black --check` clean.
- The traceback walk is robust in the edge cases I probed: a downstream `__init__` that renames `**kwargs` to `**kw` doesn't crash the walk (falls back to the enclosing frame's view), and the `isinstance(..., dict)` guard plus `from e` chaining are sound. Walking to the deepest dict-typed `kwargs` local lands exactly on the frame that forwarded to the real `object.__init__()` — precisely the final leftover set, since every intermediate mixin has already bound its named params out of it.

That closes the last open thread — no remaining suggestions. Approve.

PR #790, opened 2026-08-20
Fix #789: re-fetch backend archives unconditionally instead of gating on per-process last_*_update markers

### LINKED ISSUE #789: Mastermind doesn't pick up task/schedule changes from pyobs-robotic-backend (last_*_update markers are per-process, lost across workers)
## Summary

The Mastermind keeps running stale tasks: edits to tasks (duration, script, priority, `active`) and newly scheduled observations made in the pyobs-robotic-backend never reach the running mastermind — it keeps using the old task definitions / the old schedule.

## Root cause

pyobs-core's backend archives refresh only when a timestamp marker advances, and that marker is unreliable in the deployed multi-process setup:

1. `BackendTaskArchive._check_for_changes()` (`pyobs/robotic/storage/backend/taskarchive.py`) and `BackendObservationArchive._check_for_changes()` (`pyobs/robotic/storage/backend/observationarchive.py`) poll `GET /api/last_task_update/` resp. `/api/last_observation_update/` every 5 s and only re-download tasks/observations when the returned timestamp is newer than the locally cached `_last_update`.
2. The backend computes those timestamps purely from a Django cache marker: `post_save` receivers on `Task`/`Observation` run `cache.set("last_task_update", Time.now(), None)` (`pyobs-robotic-backend` `api/signals.py`), and the views return `cache.get("last_task_update", Time("1970-01-01T00:00:00"))` (`api/views.py`).
3. `CACHES` uses `LocMemCache` — an in-memory cache that is **per process** (`pyobs_robotic_backend/settings.py`).
4. The deployment runs **4 gunicorn workers** plus separate `celery` and `task_scheduler` processes (`docker-compose.yml`: `gunicorn -w 4 ...`). A marker set in the worker that handled a write is invisible to every other worker/process, so most polls return the `1970-01-01` fallback and `last_update` is never `> _last_update` → the archives never re-download → the mastermind runs on stale data (with keep-alive the mastermind's connection can even pin to a worker that never saw any write).

Verified: observations are returned with `task` as a plain ID (`ObservationSerializer` uses the FK), and `Observation.fetch_task()` re-resolves it from the archive's cached list on every poll — so the staleness is entirely in the archive not re-downloading, not in a stale embedded task object.

## Secondary issue (same marker mechanism, in pyobs-robotic-backend)

`mark_window_expired` and `delete_old_observations` (`api/tasks.py`) use `QuerySet.update()` / `.delete()`, which never fire `post_save` — those state changes never bump the marker even in a single-worker setup, so the mastermind keeps treating window-expired observations as pending.

## Impact

- Mastermind runs observations with outdated task parameters (duration, script, priority), including deactivated tasks (`active=False`).
- Newly scheduled observations are not picked up.
- Window-expired observations can remain runnable in the mastermind's cache.

## Suggested fixes

pyobs-core (make the archives robust against marker loss):
- Don't gate re-download solely on the marker: re-fetch unconditionally on each poll (payloads are small) or compare list contents/hash to detect changes; treat a missing/older marker as "refresh anyway".
- Don't initialize `_last_update` from the fallback timestamp.

pyobs-robotic-backend (fix the marker at the source):
- Use a shared cache backend (Redis) or compute `last_*_update` from the DB — requires adding `updated_at` to `Task`/`Observation` (doesn't exist today).
- Bump the markers in the celery bulk paths (or rewrite them to per-row saves).

## References

- pyobs-core: `pyobs/robotic/storage/backend/taskarchive.py` (`_check_for_changes`, lines 55–69), `pyobs/robotic/storage/backend/observationarchive.py` (`_check_for_changes`, lines 58–78)
- pyobs-robotic-backend: `api/signals.py`, `api/views.py` (`last_task_update`/`last_observation_update`), `api/tasks.py`, `settings.py` (`CACHES`), `docker-compose.yml` (`gunicorn -w 4`)

thusser:
## Summary

Fixes #789. The Mastermind kept running stale tasks/schedules because `BackendTaskArchive`/`BackendObservationArchive` only re-downloaded when the backend's `last_task_update`/`last_observation_update` marker advanced — but that marker is computed from a per-process Django `LocMemCache`, so under `gunicorn -w 4` most polls return the `1970-01-01` fallback, pinning `_last_update` and blocking every subsequent refresh.

## Changes

Both backend archives now **re-fetch unconditionally on every 5s poll** and detect real changes by comparing the downloaded content against the cached copy:

- **`BackendTaskArchive`** — compare `model_dump()` output (not pydantic `==`, which also compares runtime attributes like `Task._cant_run_reason` set by `can_run()`, causing spurious diffs); fire `on_tasks_changed` only on actual change.
- **`BackendObservationArchive`** — compare full `model_dump(use_task_id=True)` output **including `state`**: `Observation.__eq__` ignores `state`, so a naive equality check would miss `window_expired`/`in_progress` transitions — one of the reported symptoms. `use_task_id` keeps cached observations (whose `task` was resolved to a full `Task` by `fetch_task()`) comparable with fresh ones carrying the plain FK id.
- `_last_update` is never initialized from the backend marker anymore (no more fallback pinning); `last_changed()` reports the local time a change was last observed. `last_update_time()` stays as a public method but is no longer used for gating.

The loop body is extracted into a testable `_update()` method for both archives.

The root-cause fixes on the pyobs-robotic-backend side (shared Redis cache / `updated_at` column, marker bumps in celery bulk paths) are out of scope here and tracked separately; with this change the mastermind no longer depends on them for correctness.

## Testing

- New regression tests in `tests/robotic/storage/backend/test_backend_archives.py`: refresh not gated on the marker; no notification on unchanged content; content change detected (incl. `active` flip); runtime attributes ignored; observation **state** change detected despite `Observation.__eq__`; task-ID normalization avoids spurious updates after `fetch_task()`.
- `pytest tests/robotic` + scheduler/mastermind integration tests: 390 passed.
- `ruff`, `pyrefly`, `black` clean on touched files.

## Plan

`specs/plans/2026-08-20-backend-archive-marker-loss.md` (cross-repo: pyobs-core, pyobs-robotic-backend), added to `specs/plans/index.md`. CHANGELOG updated.

### DIFF
--- a/pyobs/robotic/storage/backend/observationarchive.py
+++ b/pyobs/robotic/storage/backend/observationarchive.py
@@ -71,16 +71,20 @@ class BackendObservationArchive(ObservationArchive):
         Re-fetches unconditionally on every poll instead of gating on the backend's
         ``last_observation_update`` marker (per-process Django ``LocMemCache``, unreliable across
         gunicorn workers -- see ``BackendTaskArchive``). Changes are detected by comparing full
-        observation contents via ``model_dump(use_task_id=True)``: the backend serializes ``task``
-        as a plain FK ID, while cached copies get their ``task`` replaced by a full ``Task`` when
-        the mastermind calls ``fetch_task()``, so the ID normalization keeps both sides
-        comparable. The comparison must include ``state`` -- ``Observation.__eq__`` ignores it, so
-        a naive list-equality check would miss e.g. window-expired or in-progress transitions.
+        observation contents via ``model_dump(use_task_id=True)``, keyed by observation ID so a
+        stable reordering of the same items is not mistaken for a change: the backend serializes
+        ``task`` as a plain FK ID, while cached copies get their ``task`` replaced by a full
+        ``Task`` when the mastermind calls ``fetch_task()``, so the ID normalization keeps both
+        sides comparable. The comparison includes ``state`` (``Observation.__eq__`` ignores it),
+        which matters for in-set transitions such as ``pending`` -> ``in_progress``; observations
+        leaving the fetched set -- e.g. ``window_expired``, which the server-side ``state``/
+        ``end_after`` filters drop -- are picked up as list shrinkage by the unconditional
+        refetch.
         """
         observations = await self._get_schedule()
-        if [o.model_dump(use_task_id=True) for o in observations] != [
-            o.model_dump(use_task_id=True) for o in self._observations
-        ]:
+        if {o.id: o.model_dump(use_task_id=True) for o in observations} != {
+            o.id: o.model_dump(use_task_id=True) for o in self._observations
+        }:
             self._observations = observations
             self._last_update = Time.now()
             sorted_obs = sorted(
@@ -245,7 +249,7 @@ class BackendObservationArchive(ObservationArchive):
             params["end_before"] = end_before.isot
         if end_after is not None:
             params["end_after"] = end_after.isot
-        observations = await http_request_paginated(self._session, url, params=params)
+        observations = await http_request_paginated(self._session, url, params=params, strict=True)
         return ObservationList([self.pyobs_model_validate(Observation, obs) for obs in observations])
 
 
--- a/pyobs/robotic/storage/backend/observationarchive.py
+++ b/pyobs/robotic/storage/backend/observationarchive.py
@@ -60,22 +60,38 @@ class BackendObservationArchive(ObservationArchive):
 
         while True:
             try:
-                last_update = await self.last_update_time()
-                if self._last_update is None or self._last_update < last_update:
-                    self._observations = await self._get_schedule()
-                    sorted_obs = sorted(
-                        filter(lambda o: o.state == ObservationState.PENDING, self._observations),
-                        key=lambda o: o.start,
-                    )
-                    if len(sorted_obs) == 0:
-                        log.info("Downloaded new schedule.")
-                    else:
-                        obs = sorted_obs[0]
-                        log.info("Downloaded new schedule. Next observation is task %s at %s.", obs.task, obs.start)
-                    self._last_update = last_update
+                await self._update()
             except Exception as e:
                 log.error("Failed to update observations from backend: %s", e)
             await asyncio.sleep(5)
+
+    async def _update(self) -> None:
+        """Fetch the schedule from the backend and apply it if anything changed.
+
+        Re-fetches unconditionally on every poll instead of gating on the backend's
+        ``last_observation_update`` marker (per-process Django ``LocMemCache``, unreliable across
+        gunicorn workers -- see ``BackendTaskArchive``). Changes are detected by comparing full
+        observation contents via ``model_dump(use_task_id=True)``: the backend serializes ``task``
+        as a plain FK ID, while cached copies get their ``task`` replaced by a full ``Task`` when
+        the mastermind calls ``fetch_task()``, so the ID normalization keeps both sides
+        comparable. The comparison must include ``state`` -- ``Observation.__eq__`` ignores it, so
+        a naive list-equality check would miss e.g. window-expired or in-progress transitions.
+        """
+        observations = await self._get_schedule()
+        if [o.model_dump(use_task_id=True) for o in observations] != [
+            o.model_dump(use_task_id=True) for o in self._observations
+        ]:
+            self._observations = observations
+            self._last_update = Time.now()
+            sorted_obs = sorted(
+                filter(lambda o: o.state == ObservationState.PENDING, self._observations),
+                key=lambda o: o.start,
+            )
+            if len(sorted_obs) == 0:
+                log.info("Downloaded new schedule.")
+            else:
+                obs = sorted_obs[0]
+                log.info("Downloaded new schedule. Next observation is task %s at %s.", obs.task, obs.start)
 
     async def last_update_time(self) -> Time:
         """Fetches last schedule update time."""
--- a/pyobs/robotic/storage/backend/taskarchive.py
+++ b/pyobs/robotic/storage/backend/taskarchive.py
@@ -70,13 +70,15 @@ class BackendTaskArchive(TaskArchive):
         mastermind run stale tasks forever. Real changes are detected by comparing the downloaded
         content against the cached copy. The comparison uses ``model_dump()`` rather than pydantic
         ``==``, which also compares runtime attributes (e.g. ``Task._cant_run_reason`` set by
-        ``can_run()``) and would flag unchanged tasks as changed on every poll.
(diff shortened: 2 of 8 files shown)


thusser (commented):
Automated review findings (3) from a code review pass.

pyobs/robotic/storage/backend/taskarchive.py
```diff
@@ -56,18 +56,34 @@ async def _check_for_changes(self) -> None:
...
+
+        Re-fetches unconditionally on every poll instead of gating on the backend's
+        ``last_task_update`` marker: that marker is computed from a per-process Django
+        ``LocMemCache`` and is unreliable across gunicorn workers, so gating on it let the
+        mastermind run stale tasks forever. Real changes are detected by comparing the downloaded
+        content against the cached copy. The comparison uses ``model_dump()`` rather than pydantic
+        ``==``, which also compares runtime attributes (e.g. ``Task._cant_run_reason`` set by
+        ``can_run()``) and would flag unchanged tasks as changed on every poll.
+        """
+        projects = await self._get_projects()
```
thusser, line 77: **Correctness — pagination truncation can silently corrupt the cache**

`_update()` now refetches unconditionally every 5s instead of only on rare marker advances. `_get_projects()`/`_get_tasks()` (and `_get_schedule()` in `observationarchive.py:80`) go through `http_request_paginated()` (`pyobs/utils/http.py`), which on a non-first-page "Invalid page." 404 (documented as happening "when concurrent writes shift page boundaries on a live-growing dataset" — exactly the multi-worker gunicorn scenario this PR targets) logs a warning and returns a **truncated** partial list instead of raising.

Neither `_update()` checks completeness before comparing/overwriting: a truncated fetch reads as "changed" and unconditionally replaces `self._tasks`/`self._projects`/`self._observations`. `get_task(id)` would then return `None` for tasks that still exist on the backend — silently, with only a WARNING log line, and now on every poll instead of rarely.

Worth a length/sanity guard (or surfacing pagination truncation as an error) before treating a shorter result as authoritative.

pyobs/robotic/storage/backend/taskarchive.py
```diff
@@ -56,18 +56,34 @@ async def _check_for_changes(self) -> None:
...
+        tasks = await self._get_tasks()
+        if [p.model_dump() for p in projects] != [p.model_dump() for p in self._projects] or [
```
thusser: **Correctness (plausible) — order-sensitive comparison risks spurious change notifications**

This is a plain Python list `!=` on `model_dump()` output, which is order-sensitive. Neither `_get_projects()`/`_get_tasks()` here nor `get_observations()` in `observationarchive.py` pass an explicit ordering param to the backend. If the backend's queryset lacks a stable `Meta.ordering` (a well-known DRF pitfall — `UnorderedObjectListWarning`), two polls 5s apart could return the same *set* of items in a different order under concurrent writes, and this comparison would report a false "changed" — firing `on_tasks_changed()` / bumping `_last_update` every poll and defeating the point of content-based change detection.

Consider comparing as a dict/set keyed by ID rather than a positional list, or requesting an explicit stable ordering.

pyobs/robotic/storage/backend/observationarchive.py
```diff
@@ -60,23 +60,39 @@ async def _check_for_changes(self) -> None:
...
+    async def _update(self) -> None:
+        """Fetch the schedule from the backend and apply it if anything changed.
+
+        Re-fetches unconditionally on every poll instead of gating on the backend's
+        ``last_observation_update`` marker (per-process Django ``LocMemCache``, unreliable across
+        gunicorn workers -- see ``BackendTaskArchive``). Changes are detected by comparing full
+        observation contents via ``model_dump(use_task_id=True)``: the backend serializes ``task``
+        as a plain FK ID, while cached copies get their ``task`` replaced by a full ``Task`` when
+        the mastermind calls ``fetch_task()``, so the ID normalization keeps both sides
+        comparable. The comparison must include ``state`` -- ``Observation.__eq__`` ignores it, so
```
thusser: **Test coverage — `window_expired` regression test doesn't reflect the real state filter**

`_get_schedule()` (line ~111) always requests `state=pending,in_progress` from the backend. So in production, an observation transitioning to `window_expired` would simply stop being returned by the API (list-length change), not come back with `state=window_expired` as `test_obs_update_detects_state_change` simulates by mocking `http_request_paginated` directly (bypassing the real filter).

That means the state-comparison logic's actual, verified value is for `pending ↔ in_progress` transitions (which stay inside the filtered set), not for `window_expired` as the docstring and plan doc claim. The test gives false confidence: it validates the comparison mechanism in the abstract but doesn't prove it's what fixes the `window_expired` symptom from #789 — the unconditional refetch + list-shrinkage does that. Worth relabeling the test/docstring claim, or adding a test that goes through the actual `state` filter path.

pyobs/robotic/storage/backend/taskarchive.py
thusser: Addressed in b7ffbace. `http_request_paginated` now has a `strict` mode (default False, so existing callers keep the tolerance) that raises on a mid-fetch "Invalid page." 404 instead of returning a partial list; `_get_projects`/`_get_tasks`/`get_observations` all pass `strict=True`, so a truncated fetch is caught by the poll loop, logged, and retried on the next poll — the cache is never replaced with a partial list. Tests: `test_paginated_strict_raises_on_invalid_page_beyond_first` in test_http.py, plus strict=True wiring assertions in the archive fetch tests.

pyobs/robotic/storage/backend/taskarchive.py
thusser: Addressed in b7ffbace. Both archives now compare ID-keyed dicts of `model_dump()` output instead of positional lists, so a stable reordering of the same items is not a change. Tests: `test_task_update_order_insensitive` and `test_obs_update_order_insensitive`. (For the record, the backend models do declare `Meta.ordering` — Task/Project by `code`, Observation by `start` — so in practice this was theoretical, but the comparison no longer depends on it.)

pyobs/robotic/storage/backend/observationarchive.py
thusser: Agreed — corrected in b7ffbace. In production the server-side `state=pending,in_progress` + `end_after=now` filters drop a window-expired observation from the response, so that symptom is fixed by the unconditional refetch applying list shrinkage; the state-comparison logic's real value is in-set transitions. The test is relabeled `test_obs_update_detects_state_transition_in_fetched_set` (pending -> in_progress), a new `test_obs_update_applies_shrinkage_when_observation_disappears` covers the window-expired production path, and the plan doc/CHANGELOG wording was corrected accordingly.

Issue #769, opened 2026-08-16
Dedupe per-frame header/JSON build in BaseVideo raw_handler for multiple raw consumers


thusser:
## Context

Flagged in code review on #766 (see [comment](https://github.com/pyobs/pyobs-core/pull/766#issuecomment-5309300616)): `_raw_frame()` runs inside each connection's own `raw_handler` task rather than once per frame and shared. With N simultaneous raw clients (guiding, a recorder, a second viewer — explicitly anticipated by `specs/design/basevideo-raw-frame-streaming.md`), the header build, JSON serialization, and `ascontiguousarray`+`tobytes()` copy all happen N times per frame for identical output.

## Why not fixed in #766

The obvious fix — precompute in `_set_image()` — conflicts with the design's "costs nothing when no raw client is connected" property (§1 of the design doc): `_set_image()` runs on every incoming frame regardless of consumers, so moving the header/JSON build there would pay that cost even with zero raw clients connected.

## Proposed fix

A lazy cache keyed by frame number: the first `raw_handler` to wake for a given frame computes and caches `(meta_bytes, frame_bytes)`; subsequent handlers waking for the same frame reuse the cached result instead of recomputing. Costs nothing when 0 or 1 raw clients are connected, dedupes work once 2+ are.

## When to pick this up

Deferred until multiple simultaneous raw consumers become a near-term reality — not needed for the single-consumer case #766 ships.

thusser:
Fixed by #807.

PR #807, opened 2026-08-24
Dedupe per-frame header/JSON build in BaseVideo raw_handler

thusser:
## Summary
- Closes #769: with N simultaneous raw clients, each connection's own `raw_handler` task independently rebuilt the header, JSON meta, and raw bytes for the same frame.
- Adds a lazy per-frame cache keyed by `_frame_num`: the first handler to wake for a frame computes and caches `(meta, frame)` bytes; later handlers waking for the same frame reuse it instead of recomputing.
- `_raw_frame()` is fully synchronous, so no other task can interleave between the cache check and the store — no lock needed.
- Costs nothing with 0 or 1 raw clients connected, matching the "costs nothing when no raw client is connected" property in `specs/design/basevideo-raw-frame-streaming.md` §1.

## Test plan
- [x] `pytest tests/modules/camera/test_basevideo.py` — 72 passed, including two new tests: dedup across two concurrently-woken consumers on the same frame, and recompute after a new frame arrives.
- [x] Full suite: 1570 passed, 25 skipped (7 pre-existing unrelated failures in `tests/cli/test_pyobsd.py`, confirmed present on `develop` without this change — environment lacks the `pyobs` executable).
- [x] `ruff check`, `black --check`, `pyrefly check` all clean on changed files.

https://claude.ai/code/session_01Fnu6xv2jcUfFKVgEbgqTae

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,10 +2,11 @@ v2.0.0.dev78 (unreleased)
 *************************
 * ``BaseVideo.raw_handler`` now caches the per-frame ``(meta, frame)`` bytes it builds
   (header assembly, JSON serialization, ``ascontiguousarray``/``tobytes()`` copy), keyed by
-  ``_frame_num``. With N simultaneous raw clients (guiding, a recorder, a viewer -- see
-  ``specs/design/basevideo-raw-frame-streaming.md``), only the first consumer to wake for a
-  given frame does the work; the rest reuse the cached result instead of redoing it (#769).
-  Costs nothing with 0 or 1 raw clients connected.
+  ``_frame_num``. With N simultaneous raw clients (guiding, a recorder, a viewer -- a case
+  the raw endpoint (``specs/design/basevideo-raw-frame-streaming.md``) already supports but
+  didn't dedupe for), only the first consumer to wake for a given frame does the work; the
+  rest reuse the cached result instead of redoing it (#769). Costs nothing with 0 or 1 raw
+  clients connected.
 * ``Application`` now logs module-creation failures (unconsumed config keys, broken ``__init__``
   chains, ...) at ERROR level with the full traceback before re-raising, so they land in the
   configured log file / journald instead of only on stderr -- which is ``/dev/null`` for
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,11 @@
 v2.0.0.dev78 (unreleased)
 *************************
+* ``BaseVideo.raw_handler`` now caches the per-frame ``(meta, frame)`` bytes it builds
+  (header assembly, JSON serialization, ``ascontiguousarray``/``tobytes()`` copy), keyed by
+  ``_frame_num``. With N simultaneous raw clients (guiding, a recorder, a viewer -- see
+  ``specs/design/basevideo-raw-frame-streaming.md``), only the first consumer to wake for a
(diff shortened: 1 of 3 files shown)


thusser (commented):
## Review: dedupe per-frame build in `raw_handler`

**Verdict: approve.** Small, correctly reasoned, well-tested. Verified the concurrency claims hold and both new tests pass; two doc-reference nits below, neither blocking.

### What changed
Caches the expensive per-frame work in `raw_handler` — header assembly, JSON serialization, and the `ascontiguousarray`/`tobytes()` copy (~8 MB for a 2048² uint16 frame) — keyed by `_frame_num`, so when N consumers wake for the same coalesced frame only one builds.

### Correctness — the reasoning checks out
- **The no-lock argument holds.** `_raw_frame()` is fully synchronous (`Image()`, `add_local_fits_headers()` — sync per `fitsheader.py:154` — `_json_safe`, `json.dumps`, `ascontiguousarray().tobytes()`), and the check → build → store block in `raw_handler` contains no `await`. Under cooperative scheduling no other task can interleave, so two consumers can never both miss for the same frame — exactly one build per wake-batch, regardless of which task runs first (even if a subclass's `activate_camera()` suspends, the first task stores the cache before its next suspension).
- **The cache key is a valid frame identity.** `_last_image` and `_frame_num` are only mutated together in `_set_image` (one synchronous block), and `_new_frame.set()` happens *after* both — so a woken consumer always reads a consistent `(frame, key)` pair; a hit can never serve a stale frame, a new frame always misses.
- **Sharing the bytes objects across consumers is safe** — immutable, and aiohttp's write path doesn't mutate the payload.
- **Memory is bounded and no worse than today** — at most one retained frame's bytes, overwritten on each miss; previously N overlapping consumers each held a transient copy.
- **CHANGELOG claims are accurate** — "only the first consumer to wake for a given frame does the work" and "costs nothing with 0 or 1 raw clients" (one int compare + tuple store per frame with a single client).

### Tests
- `test_raw_handler_dedupes_frame_build_across_consumers` is the meaningful one: both handlers genuinely parked on the event wait, one `_set_image()` wakes both, and the atomicity argument makes `raw_frame_spy.call_count == 1` deterministic regardless of ordering. The comment's justification for *not* calling `raw_handler()` with the event pre-set is correct — the first task would clear the event and the second would block on `wait()`.
- `test_raw_handler_recomputes_after_new_frame` covers the invalidation path.
- Verified on this head: **72/72 tests in `test_basevideo.py` pass** (including both new ones), `ruff check` and `black --check` clean, and the branch merges cleanly into `develop` (no conflicts; develop hasn't touched the three files since the merge base).

### Nits (non-blocking)
1. **`(design doc §769)` is a bogus reference** — the comment in `raw_handler` cites "design doc §769", but `specs/design/basevideo-raw-frame-streaming.md` only has sections §1–§5, and the dedupe cache isn't described in it at all (it's new here). Should read `(#769)` like the commit/changelog.
2. **CHANGELOG pointer slightly oversells the design doc**: "see `specs/design/basevideo-raw-frame-streaming.md`" for the caching behavior, but that doc documents the raw endpoint, not this cache. Either add a short section to the design doc covering the dedupe, or soften the reference.
3. Minor: `responses[0].write.await_args.args == responses[1].write.await_args.args` would pass even without caching (header build is deterministic per frame) — harmless sanity check, but the spy `call_count` is the real assertion.

Issue #811, opened 2026-08-24
Script/pydantic models across robotic/scripts/** have no field descriptions for pyobs-robotic-backend's form builder


thusser:
## Context

pyobs-robotic-backend's script builder renders each field's JSON-schema `description` as
help text under the form control. Every user-facing `Script` subclass (the pydantic models
that get instantiated to configure a task) is affected, not just `InstrumentConfig`.

Surveyed all of `pyobs/robotic/scripts/**` (the `Script` base is
`pyobs/robotic/scripts/script.py`): **zero fields anywhere in this tree have an explicit
`Field(description=...)`.** The only field that currently shows any help text at all is
`InstrumentConfig.image_type`, and that's a bug, not a feature -- pydantic auto-derives an
enum field's schema `description` from the enum class's own Python docstring, so
`ImageType`'s docstring ("Enumerator specifying the image type.\n\nAttributes:\n    BIAS:
...") leaks straight into the UI as one run-on line (the frontend renders it via
`textContent`, so the docstring's newlines collapse). Every other field, on every script,
shows nothing.

### `pyobs/robotic/scripts/imaging/imaging.py` (original scope of this issue)

- `InstrumentConfig`: `exposure_time`, `count`, `image_type`, `binning`, `window`,
  `optical_filter`
- `Configuration`: `acquisition_config`, `guiding_config`, `instrument_configs`, `repeats`
- `AcquisitionConfig`: `enabled`, `optional`
- `GuidingConfig`: `enabled`, `optional`
- `ImagingScript`: `configuration`, `camera`, `telescope`, `filters`, `autoguider`,
  `acquisition`

### Other `Script` subclasses

- `script.py` -- `Script` (base): `exptime_done`
- `calibration/darkbias.py` -- `DarkBiasScript`: `camera`, `count`, `exptime`, `binning`
- `calibration/pointing.py` -- `PointingScript`: `telescope`, `pointing` (nested, see below)
- `calibration/skyflats.py` -- `SkyFlatsScript`: `roof`, `telescope`, `flatfield`,
  `functions`, `priorities` (nested, see below), `min_exptime`, `max_exptime`, `timespan`,
  `filter_change`, `count`, `readout`
- `control/cases.py` -- `CasesRunner`: `expression`, `cases`
- `control/conditional.py` -- `ConditionalRunner`: `condition`, `true`, `false`
- `control/parallel.py` -- `ParallelRunner`: `scripts`, `check_all_can_run`
[...]

PR #812, opened 2026-08-25
Add Field(description=...) to all robotic Script models

thusser:
## Summary
- Every `Script` subclass under `pyobs/robotic/scripts/**` (and the nested skyflats-priority
  config models it references) is a pydantic model that pyobs-robotic-backend's script
  builder renders as a form, using each field's JSON-schema `description` as help text.
  None of them had one.
- `InstrumentConfig.image_type` was the sole exception, but only because pydantic
  auto-derives an enum field's schema `description` from the enum class's own docstring, so
  `ImageType`'s Python docstring was leaking into the UI as a run-on line.
- Adds explicit `Field(description=...)` to every field across all ~14 `Script` subclasses
  plus `ArchiveSkyflatPriorities` / `ConstSkyflatPriorities`. A field-level description
  overrides the enum-docstring leak too, so this fixes both issues in one pass.
- `pyobs/robotic/storage/lco/scripts/**` (LCO-portal-driven scripts) is intentionally left
  out, since those aren't hand-configured through the script builder form.

Closes #811

## Test plan
- [x] `ruff check pyobs/robotic/` passes
- [x] `pyrefly check` on changed files: 0 errors
- [x] `pytest tests/robotic/scripts tests/utils/skyflats tests/utils/test_config_schema.py`: 148 + 7 passed

### DIFF
--- a/pyobs/robotic/scripts/calibration/darkbias.py
+++ b/pyobs/robotic/scripts/calibration/darkbias.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING, Annotated
 
+from pydantic import Field
+
 from pyobs.interfaces import (
     IBinning,
     IData,
@@ -23,10 +25,12 @@ log = logging.getLogger(__name__)
 class DarkBiasScript(Script):
     """Script for running darks or biases."""
 
-    camera: Annotated[str, IData, IBinning, IWindow, IExposureTime, IImageType]
-    count: int = 20
-    exptime: float = 0
-    binning: tuple[int, int] = (1, 1)
+    camera: Annotated[str, IData, IBinning, IWindow, IExposureTime, IImageType] = Field(
+        description="Name of the camera module to expose with."
+    )
+    count: int = Field(default=20, description="Number of exposures to take.")
+    exptime: float = Field(default=0, description="Exposure time in seconds. 0 takes a bias, anything else a dark.")
+    binning: tuple[int, int] = Field(default=(1, 1), description="Detector binning as (x, y).")
 
     async def can_run(self, data: TaskData | None) -> bool:
         """Whether this config can currently run.
--- a/pyobs/robotic/scripts/calibration/pointing.py
+++ b/pyobs/robotic/scripts/calibration/pointing.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING, Annotated
 
+from pydantic import Field
+
 from pyobs.interfaces import IPointingAltAz, IReady
 from pyobs.robotic.scripts import Script
 from pyobs.robotic.utils.skyflats.pointing import SkyFlatsBasePointing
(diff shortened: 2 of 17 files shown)


thusser:
## Review: approve ✅

I reviewed the full diff (commit `aa527db4`) and verified the claims in the PR description.

### Coverage — complete
- All 14 `Script` subclasses under `pyobs/robotic/scripts/` accounted for: 13 got descriptions; `LcoScript` is the one intentionally excluded, matching the stated rationale.
- Both `SkyflatPriorities` subclasses covered; `SkyFlatsBasePointing` strategies have no configurable fields.
- Every field in every changed file has a description; the diff contains only `Field(...)` conversions and the `from pydantic import Field` import — defaults preserved verbatim, so this is behavior-neutral.

### Enum-docstring leak — mechanism confirmed
Verified empirically: pre-fix, `image_type`'s schema field description is `null` and the enum docstring sits on the `ImageType` `$defs` entry. The backend form builder (`schemaform.js` `resolveSchema`) merges `{...base, ...schema}` when resolving a `$ref`, so the multi-line docstring became the field's help text. The field-level description now overrides it. Diagnosis and fix are exactly right.

### Description accuracy — spot-checked against code
- `darkbias.exptime`: "0 takes a bias, anything else a dark" ↔ `ImageType.BIAS if self.exptime == 0 else ImageType.DARK`
- `cases`: `"else"` fallback key ↔ `__get_script`
- `parallel`/`sequential` `check_all_can_run` (all vs any / all vs first) ↔ `can_run` implementations
- `imaging` "Required if…" claims for telescope/filters/autoguider/acquisition ↔ `can_run` checks
- `autofocus.telescope` "…stopped afterwards" ↔ `finally: stop_motion()`
- `skyflats.readout` `"BxB"` key format ↔ scheduler `f"{binning}x{binning}"`
- `archive.binnings` "as N for NxN" ↔ `priorities[fn, (b, b)]`
- `window` "(left, top, width, height)" ↔ `IWindow.set_window` signature
- `callmodule.interface` "Full class path of the interface that declares `method`" ↔ `get_class_from_string` + `hasattr`

### Checks — all green
- `ruff check pyobs/robotic/`: pass (touched files also black/ruff-format clean)
- `pyrefly check` on changed files: 0 errors
- `pytest tests/robotic/scripts tests/utils/skyflats tests/utils/test_config_schema.py`: 155 passed

### Minor, non-blocking suggestions
1. **No regression guard** — `tests/utils/test_config_schema.py` exercises a different (dataclass-based) schema path, so nothing prevents a future PR from dropping a description. A one-liner asserting e.g. `ImagingScript.model_json_schema()["properties"]["camera"]["description"]` would lock this in.
2. **Wording nits** — `SkyFlatsScript.functions` says "Sky-brightness model functions"; the functions actually model ideal exposure time f(solar altitude) per filter/binning, and the three accepted shapes of `FlatFunctions` (plain string / dict by filter-or-binning / nested binning→filter) are opaque in the UI, so the description could mention them. `roof` "checked to be open" is actually an `IReady` check — fine as a gloss. `image_type` is accurate but terse; now that the field-level text replaces the enum docstring, enumerating the members would help users.
3. **Branch hygiene (not the PR diff)** — the local branch has a second commit `21f91059 "graphify"` (graphify-out artifacts) on top of `aa527db4`; GitHub shows only `aa527db4` for this PR, but pushing the branch as-is would pull that commit in. Consider dropping it before pushing.

thusser:
Addressed all three suggestions in `fc578211`:

1. **Regression guard** — added `tests/robotic/scripts/test_field_descriptions.py`: parametrized over all 20 models in scope, asserts every field has a non-empty `Field(description=...)`.
2. **Wording nits** — `SkyFlatsScript.functions` now describes the actual exposure-time-vs-solar-altitude expression and the three accepted dict shapes; `image_type` now enumerates the `ImageType` members.
3. **Branch hygiene** — the stray `graphify` commit was dropped and the branch force-pushed back to just `aa527db4` before this follow-up; branch head is now `fc578211` on top of that, no graphify artifacts.

`ruff`/`black`/`pyrefly` clean on changed files; `pytest tests/robotic/scripts tests/utils/skyflats`: 168 passed (20 new).

Issue #825, opened 2026-08-28
Robotic module widgets: IRobotic (executor) + IRoboticScheduler (planner) interfaces and GUI widgets


thusser:
**Repos:** pyobs-core · pyobs-gui

## Summary

Add two role-specific interfaces — `IRobotic` (executor, implemented by `Mastermind`) and `IRoboticScheduler` (planner, implemented by `Scheduler`) — plus two matching widgets in pyobs-gui, so every schedule-running robotic module can be started/stopped from the GUI and its current task / upcoming schedule inspected.

## Motivation

The robotic pipeline is made of modules that run a schedule. Today:

- **`Mastermind`** (`pyobs/modules/robotic/mastermind.py`) executes the schedule: it pulls the next `Observation` from an `ObservationArchive` and runs it via a `TaskRunner`. It implements `IAutonomous` → `IStartStop` → `IRunning`, so the only remotely visible state is a boolean `RunningState`. It emits `TaskStartedEvent` / `TaskFinishedEvent` / `TaskFailedEvent` (name + id, plus ETA on start).
- **`Scheduler`** (`pyobs/modules/robotic/scheduler.py`) plans the schedule: it computes observations into the `ObservationArchive` and implements `IStartStop` + `IRunnable` (`run()` = re-schedule now). Again, only `IRunning` on the wire.

From the GUI there is no way to control or inspect either:

- No widget exposes start/stop.
- The only robotic-related UI is a passive warning label (`mainwindow._check_warnings`) that polls `IAutonomous` clients' `IRunning` state every 5 s.
- The schedule/task data lives in module-internal child objects (`ObservationArchive` / `TaskArchive`) that are not reachable over comm. Clients can only reconstruct a partial picture from the three task events.

## Why two interfaces, not one

Both modules read and write the **same** `ObservationArchive` (Mastermind even writes the in-progress observation back to it), so exposing one shared interface on both would make the two module tabs show identical data — redundant. The genuinely role-owned data differs:

| | Mastermind (executor) | Scheduler (planner) |
|---|---|---|
| owns | the task it is executing right now (in-memory `_task`, target, obsnum, ETA) and **why a task can't run** (`TaskRunner.cant_run_reason`, tracked as `_last_cant_run_reason`) | the **plan** it computed and the action to **recompute it** (`IRunnable.run()`) |
| additionally knows | the immediate next observation it will pick up (`get_next_observation` every loop) | — |
[...]

PR #826, opened 2026-08-30
Implement IRobotic/IRoboticScheduler interfaces

thusser:
## Summary

Implements the pyobs-core side of #825: role-specific `IRobotic` (executor) / `IRoboticScheduler` (planner) interfaces, wired into `Mastermind` and `Scheduler`, per `specs/design/irobotic.md`.

- `IRobotic` (`Mastermind`): `RoboticState` pushed on `open`/`start`/`stop`, on every task transition (started/finished/failed), and whenever the next-up observation or its `cant_run_reason` actually changes — not just on transitions, so "waiting for weather" stays current while the system is idle/blocked rather than only refreshing when something starts.
- `IRoboticScheduler` (`Scheduler`): `get_schedule(limit=20)` returns pending/in-progress entries only, sorted by start time, capped at a hard server-side ceiling (`_MAX_SCHEDULE_LIMIT = 100`) regardless of the requested `limit`; raises `ValueError` on a negative `limit`. `SchedulerState.last_reschedule` published on `open`/`start`/`stop` and after each completed reschedule pass.
- `TaskStartedEvent`/`TaskFinishedEvent`/`TaskFailedEvent` now carry `obsnum` (backward compatible — optional field, defaults to `None`).
- No `running` field duplicated on either state dataclass; both interfaces inherit `IRunning`'s state via `IStartStop`.
- `RoboticTask.state` is a plain `str` (not `pyobs.robotic.observation.ObservationState`), so `pyobs.interfaces` doesn't gain an import-time dependency on `pyobs.robotic` (which itself imports `pyobs.interfaces`).

Design doc (`specs/design/irobotic.md`) updated alongside: fixed a base-class bug caught while implementing — `class Scheduler(Module, IStartStop, IRunnable, IRoboticScheduler)` is not a valid MRO (`IStartStop` can't be listed before a class that already inherits it; Python raises `TypeError: Cannot create a consistent method resolution order`, confirmed empirically). Dropped the now-redundant explicit `IStartStop`, matching the `IAutonomous`/`IRobotic` pattern already used on `Mastermind`.

Out of scope, per the design doc: the `pyobs-gui` widgets (`RoboticWidget`/`ScheduleWidget`), tracked separately in #825.

## Test plan

- [x] `ruff check` / `black --check` / `pyrefly check` clean on all touched files
- [x] New unit tests: `tests/interfaces/test_irobotic.py` (state dataclass round-trip serialization, `RoboticTask.from_observation`, `has_own_state()`)
- [x] New unit tests: `tests/modules/robotic/test_scheduler.py` (`get_schedule` filtering/sorting/limit/cap/`ValueError`, `SchedulerState` publishing)
- [x] New integration tests: `tests/integration/test_mastermind.py` (obsnum on events, `IRobotic` state on task start, `cant_run_reason` publishing)
- [x] Full existing suite green: `pytest` (1614 passed, 25 skipped, 7 pre-existing unrelated failures in `tests/cli/test_pyobsd.py` — confirmed present on `develop` before this change too, missing `pyobs` executable in this environment)
- [x] `pytest -m integration` on `tests/integration/test_mastermind.py`, `test_scheduler_mastermind.py`, `test_transit_mastermind.py` all green

https://claude.ai/code/session_01QrsaNDU9oSrXGHDQv3T2uG

### DIFF
--- a/pyobs/events/taskfailed.py
+++ b/pyobs/events/taskfailed.py
@@ -18,7 +18,7 @@ class TaskFailedEvent(Event):
         """Initializes a new task failed event.
 
         Args:
-            name: Name of task that just finished
+            name: Name of task that just failed
             id: Unique identifier for task
             obsnum: Per-night observation number of the run that just failed, e.g. "20260810-001"
         """
--- a/pyobs/events/taskfailed.py
+++ b/pyobs/events/taskfailed.py
@@ -6,6 +6,7 @@ from pyobs.events.event import Event
 class DataType(TypedDict):
     name: str
     id: Any
+    obsnum: str | None
 
 
 class TaskFailedEvent(Event):
@@ -13,15 +14,16 @@ class TaskFailedEvent(Event):
 
     __module__ = "pyobs.events"
 
-    def __init__(self, name: str, id: Any, **kwargs: Any):
+    def __init__(self, name: str, id: Any, obsnum: str | None = None, **kwargs: Any):
         """Initializes a new task failed event.
 
         Args:
             name: Name of task that just finished
             id: Unique identifier for task
+            obsnum: Per-night observation number of the run that just failed, e.g. "20260810-001"
         """
         Event.__init__(self)
-        self.data: DataType = {"name": name, "id": id}
+        self.data: DataType = {"name": name, "id": id, "obsnum": obsnum}
 
     @property
     def name(self) -> str:
@@ -30,6 +32,10 @@ class TaskFailedEvent(Event):
     @property
     def id(self) -> Any:
         return self.data["id"]
+
+    @property
+    def obsnum(self) -> str | None:
+        return self.data["obsnum"]
 
 
 __all__ = ["TaskFailedEvent"]
--- a/pyobs/events/taskfinished.py
+++ b/pyobs/events/taskfinished.py
@@ -6,6 +6,7 @@ from pyobs.events.event import Event
 class DataType(TypedDict):
     name: str
     id: Any
+    obsnum: str | None
 
 
 class TaskFinishedEvent(Event):
@@ -13,15 +14,16 @@ class TaskFinishedEvent(Event):
 
     __module__ = "pyobs.events"
 
-    def __init__(self, name: str, id: Any, **kwargs: Any):
+    def __init__(self, name: str, id: Any, obsnum: str | None = None, **kwargs: Any):
         """Initializes a new task finished event.
 
         Args:
             name: Name of task that just finished
             id: Unique identifier for task
+            obsnum: Per-night observation number of the run that just finished, e.g. "20260810-001"
         """
         Event.__init__(self)
-        self.data: DataType = {"name": name, "id": id}
+        self.data: DataType = {"name": name, "id": id, "obsnum": obsnum}
 
     @property
     def name(self) -> str:
@@ -30,6 +32,10 @@ class TaskFinishedEvent(Event):
     @property
     def id(self) -> Any:
         return self.data["id"]
+
+    @property
+    def obsnum(self) -> str | None:
+        return self.data["obsnum"]
 
 
 __all__ = ["TaskFinishedEvent"]
--- a/pyobs/events/taskstarted.py
+++ b/pyobs/events/taskstarted.py
@@ -52,7 +52,7 @@ class TaskStartedEvent(Event):
             eta = Time(d["eta"])
 
         # get obsnum
-        obsnum: str | None = d.get("obsnum")
+        obsnum: str | None = d["obsnum"] if isinstance(d.get("obsnum"), str) else None
 
         # return object
         return TaskStartedEvent(name=name, id=id, eta=eta, obsnum=obsnum)
--- a/pyobs/events/taskstarted.py
+++ b/pyobs/events/taskstarted.py
@@ -12,6 +12,7 @@ class DataType(TypedDict):
     name: str
     id: Any
     eta: str | None
+    obsnum: str | None
 
 
 class TaskStartedEvent(Event):
@@ -19,16 +20,17 @@ class TaskStartedEvent(Event):
 
     __module__ = "pyobs.events"
 
-    def __init__(self, name: str, id: Any, eta: Time | None = None, **kwargs: Any):
+    def __init__(self, name: str, id: Any, eta: Time | None = None, obsnum: str | None = None, **kwargs: Any):
         """Initializes a new task started event.
 
         Args:
             name: Name of task that just started
             id: Unique identifier for task
             eta: Predicted ETA for when the task will finish
+            obsnum: Per-night observation number assigned to this run, e.g. "20260810-001"
         """
         Event.__init__(self)
-        self.data: DataType = {"name": name, "id": id, "eta": None if eta is None else eta.isot}
+        self.data: DataType = {"name": name, "id": id, "eta": None if eta is None else eta.isot, "obsnum": obsnum}
 
     @classmethod
     def from_dict(cls, d: dict[str, Any]) -> Event:
@@ -49,8 +51,11 @@ class TaskStartedEvent(Event):
         if "eta" in d and isinstance(d["eta"], str):
             eta = Time(d["eta"])
 
+        # get obsnum
+        obsnum: str | None = d.get("obsnum")
+
         # return object
-        return TaskStartedEvent(name=name, id=id, eta=eta)
+        return TaskStartedEvent(name=name, id=id, eta=eta, obsnum=obsnum)
 
     @property
     def name(self) -> str:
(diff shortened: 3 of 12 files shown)


thusser (commented):
## Review: Approve with minor comments

Reviewed locally against `feature/irobotic-scheduler` head `17968cb` (built in a scratch copy; nothing in the working tree was modified). The design is sound, matches `specs/design/irobotic.md`, is backward compatible, and is well tested. Nothing blocking; the comments below are all minor.

### Verification results

| Check | Result |
|---|---|
| `ruff check` (10 touched files) | ✅ clean |
| `black --check` | ✅ clean |
| `pyrefly check` | ⚠️ 25 errors, **all in test files** — production code clean (see #5) |
| `tests/interfaces/test_irobotic.py` | ✅ 7/7 |
| `tests/modules/robotic/test_scheduler.py` (incl. new tests) | ✅ 49/49 |
| `tests/integration/test_mastermind.py -m integration` | ✅ 14/14 (self-contained, no live server) |
| Full suite (non-integration) | 1607 passed, 25 skipped, **14 failed — all 14 reproduce on `develop`** (7 `tests/cli/test_pyobsd.py` + 7 `test_dummysolartelescope.py`/`test_fitsheader.py`), so no regressions from this PR |
| `tests/test_events.py` | ✅ 63/63 |
| MRO claim from the design doc | ✅ confirmed: `Module, IStartStop, IRunnable, IRoboticScheduler` raises `TypeError: Cannot create a consistent method resolution order`; `Module, IRunnable, IRoboticScheduler` resolves fine |
| `has_own_state()` / interface registry | ✅ `True` for both; both registered; disco state features advertised |
| Event wire compatibility | ✅ old events without `obsnum` parse (`d.get("obsnum")` / generic `cls(**d)`); old receivers ignore the new key |
| State serializer | ✅ nested dataclass + `Time` + `None` round-trip works |

### Comments

**1. Design doc is now stale in the Mastermind section** (`specs/design/irobotic.md`)
The PR only updated the `Scheduler` section (the MRO fix — which is real and verified). The `Mastermind` section still says `cant_run_reason` "must be computed fresh at publish time… must **not** be read off `Mastermind._last_cant_run_reason`" (lines 91–97, 142–145), and the table references `_last_cant_run_reason` (line 41) — the field this PR deletes. The implementation's approach (store `_cant_run_reason` + `_next_observation` as a pair, publish on change) is reasonable — arguably better than recomputing, since the reason is bound to the actual next observation — but the doc should be updated to describe it instead of the removed field.

**2. `changed` detection misses a same-task observation change** (`pyobs/modules/robotic/mastermind.py` `_run_thread`)
`changed = reason != self._cant_run_reason or next is None or next.task.id != observation.task.id`. If the same *task* appears in two windows (or a reschedule shifts its times) and the reason is unchanged, no re-publish happens, so `next` can carry stale start/end times. Comparing observation identity or times (e.g. `self._next_observation.start != observation.start`) would cover it. Rare, but it undercuts the "waiting stays current" goal in that case.

**3. Late-start skip path never publishes** (`mastermind.py` `_run_thread`)
When a task passes `can_run` but is skipped for `can_start_late`/`allowed_late_start`, `_next_observation`/`_cant_run_reason` are cleared in memory but `_publish_robotic_state()` is never called, and the loop can keep skipping for a long time — subscribers keep the previously published state. Publishing on that path (or leaving the skipped observation visible) would close the gap.

**4. The cap test doesn't test the cap** (`tests/modules/robotic/test_scheduler.py::test_get_schedule_caps_limit_at_hard_ceiling`)
Only 5 observations are created, so `len(result) == 5` passes with or without the clamp (the comment admits this). A test with > `_MAX_SCHEDULE_LIMIT` entries would actually pin the `min(limit, 100)` behavior.

**5. "pyrefly check clean on all touched files" doesn't reproduce**
25 errors, all in the touched test files: `test_scheduler.py` (20), `test_irobotic.py` (3), `test_mastermind.py` (2). A mix of pre-existing patterns (`scheduler._comm` typed `None`, `_state_for(...) -> object`) and new assertions (`RoboticTask.start` is `Time | None`, so `r.start.isot` errors). Production code is clean. Worth either running pyrefly on `pyobs/` only, casting in the new assertions, or softening the claim in the PR body.

**6. `RoboticTask.from_observation` assumes a resolved `Task`** (`pyobs/interfaces/IRobotic.py`)
`observation.task.id`/`.name` crash if the observation carries a bare task id (portal-fetched observations can). This mirrors the pre-existing `s.task.id in removed` in `Scheduler._update_schedule`, so it's not a new regression — but `get_schedule` is a new client-facing RPC, so a defensive guard (or `fetch_task` before mapping) would make it robust for portal-backed archives.

**7. Trivial nits**
- `TaskFailedEvent.__init__` docstring still says "Name of task that just **finished**" (pre-existing copy-paste; the PR touches that docstring to add `obsnum`, so it could be fixed in passing).
- `TaskStartedEvent.from_dict` doesn't type-check `obsnum` (unlike `eta`) — `d.get("obsnum")` passes anything through.
- `mastermind.py` imports `RoboticTask`/`RoboticState` from `pyobs.interfaces.IRobotic` while also importing `IRobotic` from `pyobs.interfaces` — harmless inconsistency.
- Test coverage gaps: no test for Mastermind publishing on `open()`/`start()`/`stop()`, no test for the anti-spam dedup (unchanged → no re-publish), no `from_observation` test with a resolved target.

### Positives

- The `obsnum` events change is genuinely backward compatible (default `None`, tolerant parsing), and `Mastermind._obsnum` is correctly propagated on all three events including the failure path.
- No duplicated `running` field; the `IStartStop` inheritance is clean and the MRO fix is real (empirically confirmed).
- `_MAX_SCHEDULE_LIMIT` is a good call, and IN_PROGRESS naturally sorts first ("most imminent first").
- Both modules satisfy `Module`'s `missing_published_state` enforcement by publishing in `open()`.
- The integration tests are self-contained (mocked comm/runner) — no live server needed.

thusser:
Addressed in `be6aa273`.

**1. Design doc stale in the Mastermind section** — you're right, I only touched the `Scheduler` section when I fixed the MRO bug and left the `Mastermind`/`_last_cant_run_reason` text stale. Fixing the doc is still open — will follow up with a doc-only commit describing `_track_next_observation` (pair of `_cant_run_reason`/`_next_observation`, dedup'd by task id + start/end + reason, published from three call sites now) in place of the removed field.

**2. `changed` detection missed a same-task window change** — fixed: now also compares `start`/`end`, not just `reason`/task id. A rescheduled window on the same task republishes instead of looking unchanged.

**3. Late-start skip never published** — fixed: extracted `_track_next_observation()` (dedup + publish) and call it from both the "cannot run" branch and the late-start-skip branch. The skipped observation now stays visible as `next` (with `cant_run_reason=None`, since `TaskRunner.cant_run_reason()` was never consulted for this path — the field's contract stays scoped to what it's documented as). Test added: `test_mastermind_publishes_late_start_skip`.

**4. Cap test didn't test the cap** — fixed: now builds `_MAX_SCHEDULE_LIMIT + 10` observations and asserts the result is clamped to exactly `_MAX_SCHEDULE_LIMIT`, so it fails without the `min()`.

**5. pyrefly claim** — the `.isot`-on-`Time | None` errors in the new assertions are fixed with explicit `is not None` narrowing (matches your "casting" suggestion). Re-ran: those are gone; what's left in the touched test files is exactly the pre-existing `_comm: Comm | None`/mock-typing noise you already characterized as unrelated (verified — same error count on `develop` for that pattern). Production code (`pyobs/`) stays 0 errors.

**6. `from_observation` assumes a resolved task** — this was real and would have crashed `get_schedule()` on `PortalObservationArchive`: its `get_schedule()` (unlike `get_next_observation`/`get_current_observation`) returns `task` as a bare FK id, never calls `fetch_task()`. Fixed: `Scheduler.get_schedule()` now calls `obs.fetch_task(self._task_archive)` per entry before mapping, skipping (with a log line, matching the existing pattern in `get_next_observation`/`get_current_observation`) anything that fails to resolve. Tests added for both the resolve and the skip-on-failure path.

**7. Nits** — all fixed: `TaskFailedEvent` docstring, `obsnum` type-checked in `TaskStartedEvent.from_dict` (mirrors `eta`), imports consolidated to `pyobs.interfaces` top-level. Coverage gaps closed: open/start/stop publish, anti-spam dedup actually suppressing a republish, `from_observation` with a resolved target.

Full suite: 1617 passed (up from 1607 — 10 new tests), same 7 pre-existing `test_pyobsd.py` failures (still unrelated, still reproduce on `develop`). Point 1 (doc) is the only one not yet closed.

thusser:
Point 1 (doc) also closed now, in `74eaa27b` — the ownership table, `IRobotic` section, and `Mastermind` implementation section now describe `_track_next_observation`/the `_cant_run_reason`+`_next_observation` pair instead of the removed `_last_cant_run_reason` dict, and the checklist is marked done with a link back to this PR.

PR #827, opened 2026-08-30
Add DummyMastermind/DummyScheduler for pyobs-gui development

thusser:
## Summary

Follow-up to #826, for #825: adds `DummyMastermind`/`DummyScheduler`, standalone `IRobotic`/`IRoboticScheduler` implementations with no real `ObservationArchive`/`TaskArchive`/`TaskScheduler`/`TaskRunner` backing, so pyobs-gui has something to develop `RoboticWidget`/`ScheduleWidget` against without standing up the real robotic pipeline. Follows the existing `Dummy*`/`Mock*` module pattern (`DummyAcquisition`, `MockWeather`) — same domain package (`pyobs/modules/robotic/`), same "simulate over time" style as `DummyAcquisition`'s fake convergence loop.

- **`DummyMastermind`**: a background loop generates synthetic tasks, runs each for a simulated duration, and occasionally simulates a "can't run" wait beforehand or a failed run — so `RoboticWidget` has real state transitions (`current`/`next`/`cant_run_reason`, `TaskStarted`/`Finished`/`FailedEvent`) to develop against, not just a static snapshot.
- **`DummyScheduler`**: generates a contiguous synthetic schedule on `open()` and whenever `run()` is called ("Re-schedule now"), so `ScheduleWidget`'s table and re-schedule button have real data and a visible effect. `get_schedule()` derives pending/in-progress per entry from wall-clock time and drops elapsed entries, same trim/cap/`ValueError` contract as the real `Scheduler`.

The two are intentionally independent — no shared synthetic data, since they'd be independent processes in production too.

## Test plan

- [x] `ruff check` / `black --check` / `pyrefly check` clean
- [x] New tests: `tests/modules/robotic/test_dummymastermind.py` (open/start/stop publish, task start→finish cycle, simulated blocked reason, simulated failure, doesn't pick up a new task while stopped)
- [x] New tests: `tests/modules/robotic/test_dummyscheduler.py` (open/start/stop publish, `run()` triggers reschedule, schedule generation is contiguous/sorted/unique-id, `get_schedule` limit/cap/negative-limit/elapsed-filtering/in-progress-marking)
- [x] Both modules manually smoke-tested end-to-end (real `open()`/background loop/`get_schedule()` against a mocked `Comm`)
- [x] Full suite green: 1641 passed (up from 1617 — 24 new tests), same 7 pre-existing unrelated `test_pyobsd.py` failures

https://claude.ai/code/session_01QrsaNDU9oSrXGHDQv3T2uG

### DIFF
--- a/pyobs/modules/robotic/__init__.py
+++ b/pyobs/modules/robotic/__init__.py
@@ -5,9 +5,11 @@ TODO: write doc
 
 __title__ = "Robotic mode"
 
+from .dummymastermind import DummyMastermind
+from .dummyscheduler import DummyScheduler
 from .mastermind import Mastermind
 from .pointing import PointingSeries
 from .scheduler import Scheduler
 from .scriptrunner import ScriptRunner
 
-__all__ = ["Mastermind", "PointingSeries", "Scheduler", "ScriptRunner"]
+__all__ = ["DummyMastermind", "DummyScheduler", "Mastermind", "PointingSeries", "Scheduler", "ScriptRunner"]
--- a/pyobs/modules/robotic/dummymastermind.py
+++ b/pyobs/modules/robotic/dummymastermind.py
@@ -24,6 +24,9 @@ class DummyMastermind(Module, IAutonomous, IRobotic):
     "can't run" wait beforehand and an occasional failure, so `RoboticWidget` has real state
     transitions to develop against without needing a real `ObservationArchive`/`TaskRunner`.
     Independent of `DummyScheduler` -- the two don't share any synthetic data.
+
+    `RoboticState.next` is only populated during a simulated "can't run" wait, same as the real
+    `Mastermind` -- raise `blocked_probability` to exercise the next-up panel more often.
     """
 
     __module__ = "pyobs.modules.robotic"
@@ -65,6 +68,8 @@ class DummyMastermind(Module, IAutonomous, IRobotic):
         self._running = False
         self._task_ids = itertools.count(1)
         self._task: RoboticTask | None = None  # currently running
+        self._obsnum_night: str | None = None
+        self._obsnum_counter = 0
 
         self.add_background_task(self._run_thread, True)
 
@@ -93,6 +98,16 @@ class DummyMastermind(Module, IAutonomous, IRobotic):
         self._running = False
         await self.comm.set_state(IRunning, RunningState(running=self._running))
         await self._publish()
+
+    def _next_obsnum(self) -> str:
+        """Per-night observation counter, e.g. "20260810-001" -- in-memory only, unlike the
+        real Mastermind's persisted cache; a dummy doesn't need to survive restarts."""
+        night = Time.now().datetime.strftime("%Y%m%d")
+        if night != self._obsnum_night:
+            self._obsnum_night = night
+            self._obsnum_counter = 0
+        self._obsnum_counter += 1
+        return f"{night}-{self._obsnum_counter:03d}"
 
     def _make_task(self) -> RoboticTask:
         return RoboticTask(
@@ -132,10 +147,13 @@ class DummyMastermind(Module, IAutonomous, IRobotic):
             next_task.start = start
             next_task.end = end
             next_task.state = "in_progress"
+            next_task.obsnum = self._next_obsnum()
             self._task = next_task
 
             log.info("Running task %s...", next_task.name)
-            await self.comm.send_event(TaskStartedEvent(name=next_task.name, id=next_task.id, eta=end))
+            await self.comm.send_event(
+                TaskStartedEvent(name=next_task.name, id=next_task.id, eta=end, obsnum=next_task.obsnum)
+            )
             await self._publish()
 
             await asyncio.sleep(duration)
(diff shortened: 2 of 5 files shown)


thusser (commented):
## Review: Approve

A clean, well-scoped follow-up to #826. Properly stacked on the now-merged `develop` (parent of the head commit is the #826 merge `d852c01a`, so the diff is exactly the 5 new files), no dependency entanglement. Nothing blocking.

### Verification (PR head `30b198a6`, scratch copy)

| Check | Result |
|---|---|
| `ruff check` (5 files) | ✅ clean |
| `black --check` | ✅ clean |
| `pyrefly check` | ⚠️ 47 errors — **all in the two new test files** (same `_comm`/`_state_for` typing noise as #826); both new modules and `__init__.py` are clean (see #5) |
| New tests (`test_dummymastermind.py` + `test_dummyscheduler.py`) | ✅ 22/22 pass |
| Full suite (non-integration) | 1634 passed, 25 skipped, **14 failed — all 14 reproduce on `origin/develop`** (7 `test_pyobsd.py` + 7 telescope/fitsheader), so zero regressions from this PR |
| Instantiation / abstract completeness | ✅ both classes concrete; `interfaces` include `IRobotic` / `IRoboticScheduler`; state published in `open()` satisfies `missing_published_state` |
| `get_schedule` contract | ✅ limit, `_MAX_SCHEDULE_LIMIT=100` cap, `ValueError` on negative limit, elapsed-drop, in-progress derivation all verified |

### Positives

- **The cap test actually tests the cap** (`test_get_schedule_caps_limit_at_hard_ceiling` uses `schedule_size = _MAX_SCHEDULE_LIMIT + 10` and asserts the clamp) — a nice improvement over the weak version flagged in #826.
- Tests are deterministic: probabilities are forced to 0/1, so the `random` calls can't make them flaky.
- `get_schedule` faithfully mirrors the real `Scheduler` contract (default limit 20, hard ceiling, `ValueError`, most-imminent-first ordering via the contiguous cursor generation).
- Well-documented classes and config params; the intentional independence of the two modules (no shared synthetic data) is a sensible call given they'd be separate processes.
- The dummy mastermind's state machine matches the real one's semantics: `current` during a run, `next`+`cant_run_reason` only when blocked, cleared after finish — so the widget sees realistic transitions.

### Comments (minor)

**1. Task events don't carry `obsnum`** (`pyobs/modules/robotic/dummymastermind.py`)
`TaskStartedEvent`/`FinishedEvent`/`FailedEvent` are sent without the `obsnum` #826 added. It's optional so this is wire-compatible, but since the whole point is giving `RoboticWidget` real data to develop against, a synthetic per-night obsnum (e.g. `f"{now:%Y%m%d}-{counter:03d}"` from the existing `itertools.count`) would let the widget's obsnum field be exercised too.

**2. `next` is only ever populated during a simulated blocked wait**
In the normal flow the loop makes a task and immediately starts it, so `RoboticState.next` is `None` except during blocked periods. That mirrors the real Mastermind exactly, so it's fine — but GUI developers will see an empty "next-up" panel most of the time unless they raise `blocked_probability`. Worth a sentence in the docstring.

**3. `get_schedule` mutates the stored schedule in place** (`pyobs/modules/robotic/dummyscheduler.py`)
`task.state = "in_progress"/"pending"` is written back onto the internal `RoboticTask` objects on every call. Harmless for a dummy, but the in-memory schedule isn't a stable snapshot if a caller holds references. Trivial.

**4. First-call race after `open()`**
The schedule is generated on the worker's first tick, so a `get_schedule` call in that window returns `[]` (and `last_reschedule` is `None` until then). The real Scheduler has the same transient, so it's consistent — but generating synchronously in `open()` would remove it for widget dev.

**5. "pyrefly check clean" only holds for source files**
47 errors, all in the new test files (the `_comm`-typed-`None` and `_state_for(...) -> object` patterns, same as #826). Production code is clean. Suggest either scoping the claim to `pyobs/` or silencing the test-typing noise.

**6. Trivial**
- `DummyScheduler.get_schedule` relies on generation order for sortedness; the real `Scheduler` sorts explicitly. A defensive `sorted()` would make it robust to manual edits of `_schedule` (tests included).
- Same `limit`-non-int nit as #826: a non-int `limit` raises `TypeError` rather than the documented `ValueError`.
- `test_run_thread_simulates_failure` asserts the `TaskFailedEvent` but not the post-failure state publish (`current=None`) — the finished-path test covers the clearing, so this is fine, just noting coverage.

thusser:
Addressed in `f4dcc6b1`.

**1. obsnum** — added: `DummyMastermind._next_obsnum()` mirrors the real Mastermind's `<night>-<counter>` scheme (in-memory only, no persisted cache — a dummy doesn't need to survive restarts). Carried on all three task events now. Tests added for both the per-night increment and the night-rollover reset.

**2. `next` only populated during a blocked wait** — agreed it's correct (mirrors the real Mastermind), added the docstring sentence you suggested.

**3. `get_schedule` mutating stored tasks** — fixed: now returns `dataclasses.replace(task, state=...)` copies instead of writing back onto `self._schedule`'s objects. Test added (`test_get_schedule_does_not_mutate_stored_tasks`).

**4. First-call race after `open()`** — fixed: `open()` now generates the first schedule synchronously (and sets `_need_update = False` so the worker's first tick doesn't immediately redo it), so `get_schedule()` never returns empty right after startup. Updated `test_open_publishes_state` to match (was asserting the now-removed transient).

**5. pyrefly** — same situation as #826: production code stays clean, the 47 errors are all the pre-existing `_comm`/`_state_for` mock-typing pattern in the two test files, not something introduced by this PR's logic. Not chasing that class of noise further, same call as last time.

**6. Trivial**:
- Added a defensive `sorted(self._schedule, key=...)` in `get_schedule` — no longer relies on generation order.
- `limit` now validated with `isinstance` before the negativity check, so a non-int raises the documented `ValueError` instead of a bare `TypeError`. (Left the real `Scheduler.get_schedule` in #826 as-is — same nit applies there, but that's a separate already-merged PR; happy to send a one-line follow-up for it if you want.)
- Left the failure-path state-publish assertion as noted — you flagged it as already covered by the finished-path test, agreed.

Full suite still green: 1645 passed (up from 1634 — 4 new tests), same 7 pre-existing unrelated `test_pyobsd.py` failures.

Issue #824, opened 2026-08-28
_retry_delay overflows after 1024 attempts and permanently kills the event subscription


MichaelJEvan:
**Versions:** pyobs-core 2.0.2, Python 3.11, macOS

`_retry_delay` (`pyobs/comm/xmpp/xmppcomm.py:59`) applies the cap *after* computing the
exponent, so the cap never protects it:

```python
return random.uniform(0, min(cap, base * (2**attempt)))
```

```python
min(30.0, 1.0 * (2**1000))   # -> 30.0
min(30.0, 1.0 * (2**1024))   # -> OverflowError: int too large to convert to float
```

`_subscribe_event_with_retry` retries indefinitely against a node that does not exist yet, so
it always reaches 1024 eventually and the task dies:

```
File ".../xmppcomm.py", line 958, in _subscribe_event_with_retry
    await self._safe_send(self.client.plugin["xep_0060"].subscribe, self._pubsub_service, node)
slixmpp.exceptions.IqError: ... <item-not-found/> Node not found

During handling of the above exception, another exception occurred:
File ".../xmppcomm.py", line 968, in _subscribe_event_with_retry
    await asyncio.sleep(_retry_delay(attempt))
File ".../xmppcomm.py", line 60, in _retry_delay
    return random.uniform(0, min(cap, base * (2**attempt)))
OverflowError: int too large to convert to float
```

The subscription is then lost permanently, not just paused. The key is added to
`_event_subscriptions` before the loop and is not removed when the task dies, so the guard at
the top of `_subscribe_event_with_retry` makes every later re-registration return immediately.
The pair stays marked as subscribed while nothing is subscribed and nothing is retrying.

Observed on a client left running overnight: started 22:47, first failure 02:57 (~4.2 hours,
about 1024 attempts at a 30 s cap with full jitter). Three subscriptions died in four minutes:
`camera:NewImageEvent`, `telescope:FilterChangedEvent`, `telescope:OffsetsRaDecEvent`.

The practical case: a client starts in the afternoon and subscribes to `NewImageEvent`. The
camera has not exposed yet, so the node does not exist and it retries. Four hours later the
task dies and the pair is permanently marked as subscribed. Night falls, the camera takes its
first image, and nothing hears about it.

Clamping the exponent before the multiply fixes the overflow; discarding the key when the task
exits abnormally would stop a dead subscription from blocking re-registration.

thusser:
[...]

PR #828, opened 2026-08-30
Pre-create pubsub event nodes at startup, harden retry machinery (#824)

thusser:
## Summary
- `XmppComm._register_events` now pre-creates a module's own event pubsub node the moment it declares a send-only event (`register_event()` with no handler), instead of relying on the server's lazy auto-create-on-first-publish. Closes the window where a peer connecting before the module's first publish could never complete a subscribe (`item-not-found`) and had to fall back to an indefinitely retrying background task.
- State nodes are **explicitly out of scope** — see the plan's "Why state nodes are excluded" section (enforce-state-publishing already narrows that gap to sub-second, and the natural code anchor for state pre-creation sits before the XMPP connection is actually established).
- Fixes #824: `_retry_delay`'s exponential backoff overflowed (`OverflowError`) at high attempt counts (e.g. attempt 1024); clamped. Both the event and state subscribe-retry background tasks now discard their tracked key/handler on an unexpected failure instead of leaving a permanently stuck "subscribed" marker behind with nothing actually subscribed and nothing retrying.
- Also fixes a pre-existing integration-test-harness race: `tests/integration/conftest.py` reconnected the same fixed XMPP resource before ejabberd had reaped the previous test's session (a connection-level kick, not an `IqError`/`IqTimeout`) — `_create_node`'s new IQ traffic was simply the first thing in the suite to actually surface it.

Plan: `specs/plans/2026-08-28-precreate-pubsub-nodes.md`.

## Status / caveats
- Unit suite (1614 tests, excluding 7 pre-existing, unrelated `test_pyobsd.py` failures caused by a local-environment `pyobs` executable path issue, confirmed present on `develop` too) is green.
- `ruff` and `pyrefly` are clean on all touched files.
- The live-ejabberd integration suite (`tests/integration/test_xmpp_event_subscriptions.py`) is still being confirmed on a freshly-started ejabberd container as of this PR — an earlier run against a container that had accumulated state from repeated prior runs saw 13/14 tests hang on connection-level issues unrelated to this change's logic. Will update this PR with results once that run completes.

## Test plan
- [x] `pytest tests/comm/test_pubsub_node_precreation.py` (14 new unit tests, no live server)
- [x] `ruff check` / `black --check` / `pyrefly check` on all touched files
- [x] Full unit suite (`pytest -m "not integration and not xmpp"`)
- [ ] Live-ejabberd integration suite on a clean container (`tests/integration/test_xmpp_event_subscriptions.py`) — in progress, will report back

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,4 @@
-v2.0.1.dev1 (2026-08-30)
+v2.0.3.dev1 (2026-08-30)
 *************************
 * ``XmppComm`` now pre-creates a module's own event pubsub nodes at the moment it declares them
   (send-only ``register_event()``), instead of relying on the server's lazy auto-create-on-first-
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,16 @@
+v2.0.1.dev1 (2026-08-30)
+*************************
+* ``XmppComm`` now pre-creates a module's own event pubsub nodes at the moment it declares them
+  (send-only ``register_event()``), instead of relying on the server's lazy auto-create-on-first-
+  publish. Closes the window where a peer connecting before a module's first publish could never
+  complete a subscribe (``item-not-found``) and had to fall back to an indefinitely retrying
+  background task. State nodes are explicitly out of scope (see
+  ``specs/plans/2026-08-28-precreate-pubsub-nodes.md``'s "Why state nodes are excluded" section).
+* Fixes issue #824: ``_retry_delay``'s exponential backoff no longer raises ``OverflowError`` at
+  high attempt counts (e.g. attempt 1024), and both event/state subscribe-retry background tasks
+  now discard their tracked key/handler on an unexpected failure instead of leaving a permanently
+  stuck "subscribed" marker behind with nothing actually subscribed and nothing retrying.
+
 v2.0.0 (2026-08-26)
 *******************
 
--- a/pyobs/comm/xmpp/xmppclient.py
+++ b/pyobs/comm/xmpp/xmppclient.py
@@ -4,7 +4,6 @@ from typing import Any
 
 import slixmpp
 import slixmpp.xmlstream
-from slixmpp.xmlstream import StanzaBase
 
 from pyobs.comm.xmpp.xep_0009.rpc import XEP_0009
 from pyobs.comm.xmpp.xep_0009_timeout import XEP_0009_timeout
@@ -71,7 +70,6 @@ class XmppClient(slixmpp.ClientXMPP):
         self.add_event_handler("failed_auth", lambda ev: self._auth(False))
         self.add_event_handler("failed_all_auth", self.failed_all_auth)
         self.add_event_handler("stream_error", self._stream_error)
-        self.add_filter("in", self._filter_messages)
 
     def send_presence(self, *args: Any, **kwargs: Any) -> Any:
         """Override slixmpp's send_presence to hold back every presence broadcast -- the initial
@@ -111,13 +109,6 @@ class XmppClient(slixmpp.ClientXMPP):
         forever. Leaving reconnection solely to XmppComm avoids that.
         """
         return self.disconnect(0.0, reason=reason)
-
-    def _filter_messages(self, stanza: StanzaBase) -> StanzaBase | None:
-        # if a user with same JID is already connected, we get a conflict
-        if '<conflict xmlns="urn:ietf:params:xml:ns:xmpp-stanzas" />' in str(stanza):
-            self._jid_conflict = True
-            return None
-        return stanza
 
     def _stream_error(self, error: Any) -> None:
         """Called when the server sends a <stream:error/>, e.g. when this connection gets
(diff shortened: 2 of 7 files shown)


thusser:
## Update: real root cause found and fixed, integration suite confirmed green

The earlier "13/14 tests hang" result was **not** a container-state artifact as I first suspected (I initially thought it was cross-session collision with another Claude Code session sharing the default `xmpp-ejabberd-1` docker-compose container/port — that turned out to be a red herring; the same failure reproduced identically on a fully isolated container too).

**Actual root cause**: `XmppClient._filter_messages` (pre-existing, predates this PR) does a naive substring search for `<conflict xmlns="urn:ietf:params:xml:ns:xmpp-stanzas" />` across *every* incoming stanza and silently drops any match. That condition string is the generic XMPP "cancel" error used for many unrelated things — including XEP-0060's "node already exists" reply, which is exactly what `_create_node`'s restart-tolerance produces and depends on receiving. The filter was written long ago to detect a JID/resource-conflict *stream* error, but a later commit (`7adbcd94`) added the correct, precise `_stream_error`/`stream_error`-event handler for that exact purpose, leaving `_filter_messages` fully redundant — and actively harmful, since it can't tell a stream-level session kick apart from an ordinary application-level `<conflict/>` IQ error.

Confirmed via a live-ejabberd repro (bisecting a real `XmppComm.open()` hang down to this one line): with the filter registered, any `create_node` hitting `<conflict/>` (i.e., every module restart where the node already exists) hangs `_safe_send` for its full retry budget instead of degrading gracefully, exactly the failure mode I was seeing. Removed the filter (last commit); `_stream_error` alone already covers the real scenario.

**Current status**, isolated container (unique compose project/port, not the shared one another session might be using):
- 14/14 integration tests pass in isolation.
- 12/14 pass in one full sequential run against a never-reset container; the 2 failures (`test_non_subscriber_never_gets_the_wire_message`, `test_unregister_stops_delivery`) are pre-existing `persist_items`/`send_last_published_item: on_sub_and_presence` cross-test pollution — confirmed by re-running just those two on a fresh container, where both pass. Not a regression from this change; a known test-harness limitation (nodes/items persist server-side across the whole file's sequential test run) that predates this PR.
- Full unit suite: 1614 passed (7 pre-existing, unrelated `test_pyobsd.py` failures confirmed present on `develop` too, from a local `pyobs`-executable-path issue).
- ruff / black / pyrefly clean on all touched files.

Plan doc (`specs/plans/2026-08-28-precreate-pubsub-nodes.md`) updated to closed/implemented status with the checklist filled in.

thusser (commented):
Reviewed the final head (`7145766b`, incl. `6b267a68` filter removal). `tests/comm/` is green locally (180 passed, including the 14 new unit tests).

The `_filter_messages` removal is safe and a genuinely important catch: `_stream_error` (`xmppclient.py`) already covers the same-JID kick path and `wait_connect()` still polls `_jid_conflict`; the old substring filter never even matched stream errors (whitespace mismatch in the search string), yet it silently dropped every stanza containing that element — including XEP-0060's "node already exists" IQ error, which `_create_node`'s restart tolerance depends on being delivered. Without this, any restart where the event node already existed would hang `open()` for the full `_safe_send` budget.

One substantive point (worth addressing, even as a follow-up):

**`_subscribe_with_retry`'s teardown wraps the fetch-and-dispatch phase, not just the subscribe loop.** The new outer `except Exception` pops the `_state_node_handlers` entry on *any* unexpected exception — including one from `_xml_to_dataclass` (a malformed item) or from a user callback (`cb(state_obj)`, e.g. a GUI widget) during the initial `get_items` dispatch. Before this change those exceptions only killed the background task (noisy "never retrieved" warning) while the subscription and all other callbacks survived; now a single bad payload or one raising callback permanently drops the node handler. Since the server-side subscription is still live, subsequent notifications then arrive and are silently discarded in `_handle_event_sync` (`node not in _state_node_handlers`) — the same class of permanent silent loss #824 fixes, just via a different trigger. Suggest scoping the discard/`raise` to failures *inside the subscribe loop* and leaving the dispatch phase as before (catch `IqError`/`IqTimeout`, propagate everything else without touching `_state_node_handlers`). The event path is clean — no dispatch code inside its try.

Minor points:

- Double logging on the state path: `log.exception(...)` inside the handler plus the `_log_task_exception` done-callback `log.error(...)` after the re-raise print the same traceback twice.
- `min(attempt, 60)` in `_retry_delay` is a magic number — with the defaults the cap is already hit at attempt 5; 60 only guards the `2**1024` float overflow. A one-line comment would help.
- Startup cost: each `_create_node` runs sequentially and can burn up to ~90 s (`5 × 15 s` timeout + jittered sleeps) against an unresponsive server, × the number of send-only events, inside `open()`. Probably acceptable (connect fails anyway), but worth a conscious note.
- CHANGELOG header `v2.0.1.dev1` vs `pyproject.toml`'s `2.0.2` on the base — if the convention is header-matches-pyproject, this should be `v2.0.2.dev1` (or `v2.0.3.dev1`).
- `conftest.py`'s fixed 2 s teardown sleep is a heuristic that adds ~2 s to every integration test; polling until the old session is actually reaped (or using a unique resource per test) would be more robust. The repro write-up is good documentation either way.
- Drive-by nit: `_safe_send`'s docstring (`xmppcomm.py` ~1061) has real except-clause text merged into it, and the method body indeed has no `except IqError` (which is why every caller catches it themselves). Pre-existing, but easy to tidy in this PR.

Not blockers: the pre-creation scope edges (events declared with a handler that the module also sends don't get their own node pre-created; `_get_derived_events` pre-creates nodes for derived events that may never be published) are documented in the plan and benign in effect.

Issue #830, opened 2026-08-31
http_request_with_retries: only warn on a failed connection after a minute or so, not on every retry


thusser:
`http_request_with_retries` (`pyobs/utils/http.py`) is decorated with tenacity's `@retry`, using
`before_sleep_log(log, logging.WARNING)` and `wait_exponential(multiplier=1, min=1, max=30)` with 5
attempts. Every failed attempt therefore logs a WARNING like

```
Retrying pyobs.utils.http.http_request_with_retries in 2 seconds as it raised ClientConnectorError:
Cannot connect to host ...
```

The periodic pollers in `pyobs/robotic/storage/portal/` (taskarchive, observationarchive) run every
few seconds, so while an HTTP endpoint is unreachable the log fills with near-identical
"Retrying ..." WARNING lines every couple of seconds — on top of the final ERROR each poll cycle
already logs when the request ultimately fails after all attempts. One outage produces a wall of
retry warnings in which the actual one-line error is buried.

Requested behavior: do not warn on every failed attempt; only warn once the connection has been
failing for roughly a minute or so.

- Retry-attempt messages should be silent or DEBUG during the first ~60 s of failure.
- Once an endpoint has been failing continuously for ~60 s, emit a WARNING — and keep it sparse
  afterwards (one per ~minute, not one per retry).
- Keep the final ERROR after all attempts are exhausted (callers already log it, so the per-retry
  WARNING adds no information).

Implementation notes:

- The current decorator is static (`before_sleep=before_sleep_log(log, logging.WARNING)`). Options
  are a custom `before_sleep` callback that picks the level from the elapsed time since the first
  failure, or a small stateful tracker that records the first-failure time per URL — the latter
  also spans successive requests, not just the retries within a single call.
- "A minute" is a policy choice; the point is that the first warning should come far later than the
  first retry (backoff alone only spans ~15 s per request, and pollers repeat every few seconds).

thusser:
Fixed in #835 (merged `ce924630`): `http_request_with_retries` now stays quiet for the first ~60s a URL has been failing, then warns at most once/minute until it recovers. Also extended the same throttling to `PortalTaskArchive`/`PortalObservationArchive`'s poll-loop ERROR logging (first failure logs immediately, repeats during the same outage throttle to once/minute) since that would otherwise have kept spamming at ERROR level even with the WARNING fix in place.

PR #835, opened 2026-08-31
Throttle retry-warning spam in http_request_with_retries

thusser:
## Summary
- Closes #830. `http_request_with_retries` (`pyobs/utils/http.py`) previously logged a WARNING on every single failed retry attempt via a static `before_sleep_log`. During a pyobs-portal restart (typically offline ~1-2 min), the pollers in `pyobs/robotic/storage/portal/` flooded the log with near-identical retry warnings.
- Replaced the static `before_sleep` with a custom callback backed by a per-URL `(first_failure_time, last_warned_time)` tracker: retries stay at DEBUG for the first ~60s a URL has been failing, then warn at most once a minute. A successful request clears the tracked state, so the next restart starts its own clean 60s window.
- The final ERROR after all attempts are exhausted (already logged by callers) is untouched.

## Test plan
- [x] `pytest tests/utils/test_http.py` (19 tests, including new coverage for: quiet-before-threshold, single warning past threshold, throttled repeat warnings, and state cleared on success)
- [x] `ruff check` / `black --check`
- [x] `pyrefly check pyobs/utils/http.py`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,10 @@ v2.1.1.dev1 (2026-08-31)
 * Fixes issue #830: ``http_request_with_retries`` no longer logs a WARNING on every failed retry
   attempt. Retries now stay quiet for the first 60s a URL has been failing (covering a typical
   short pyobs-portal restart), then warn at most once per minute until the request succeeds again.
+  ``PortalTaskArchive``/``PortalObservationArchive``'s poll loops got the same treatment for their
+  ``log.error`` calls: the first failure still logs immediately at ERROR, but repeats during the
+  same outage are throttled to at most once per minute instead of once per ~20s poll cycle. Both
+  reuse a new ``pyobs.utils.http.LogThrottle`` helper.
 
 v2.1.0 (2026-08-31)
 ********************
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+v2.1.1.dev1 (2026-08-31)
+*************************
+* Fixes issue #830: ``http_request_with_retries`` no longer logs a WARNING on every failed retry
+  attempt. Retries now stay quiet for the first 60s a URL has been failing (covering a typical
+  short pyobs-portal restart), then warn at most once per minute until the request succeeds again.
+
 v2.1.0 (2026-08-31)
 ********************
 * Added ``IRobotic`` (executor) / ``IRoboticScheduler`` (planner) interfaces, wired into
--- a/pyobs/robotic/storage/portal/observationarchive.py
+++ b/pyobs/robotic/storage/portal/observationarchive.py
@@ -9,7 +9,7 @@ import aiohttp
 
 from pyobs.robotic import ObservationArchive, Task, TaskArchive
 from pyobs.robotic.observation import Observation, ObservationList, ObservationState
-from pyobs.utils.http import http_request_paginated, http_request_with_retries
+from pyobs.utils.http import LogThrottle, http_request_paginated, http_request_with_retries
 from pyobs.utils.time import Time
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


thusser:
## Review: looks good — approve with two decisions to make

Verified locally: `pytest tests/utils/test_http.py` (19 passed), `ruff` / `black` clean, `pyrefly check pyobs/utils/http.py` (0 errors). I also confirmed empirically with tenacity 9.1.4 that with `stop_after_attempt(5)` and waits 1/2/4/8, `before_sleep` fires 4× per call (after attempts 1–4, never the last) and a single call spans only ~15 s — so the cross-call `_failure_state` is what actually makes the throttle work, and it does.

The one thing I double-checked hard: **the URL keys are stable in the real poller scenario**. `observationarchive.py` / `taskarchive.py` pass the base URL as the `url` positional arg; time-varying values (`end_after=Time.now()`, `state=…`) go in `params=`, never in the URL string. So the throttle holds across polls. Good — that is the single thing that would silently break this design, and it does not here.

### 1. Caller-level ERROR spam remains (please make this a stated decision)
Both portal archives still `log.error("Failed to update observations/tasks from portal: …")` every poll cycle (~1 per ~20 s per archive during an outage). The PR body acknowledges it, and the issue anticipated it — but the outcome is still a noisy log during outages, just at ERROR level. If #830’s goal is "don’t bury the one-line error," consider throttling the caller loop too (first failure at ERROR, subsequent at DEBUG until the URL recovers — `_failure_state` already has the machinery). At minimum, state this as a deliberate scope cut in the PR.

### 2. Changelog version/date
`v2.1.1.dev1 (2026-08-31)` while `pyproject.toml` still says `2.1.0`, and the dev entry shares the v2.1.0 release date. If changelog-first is the convention, fine — just confirm the version bump belongs in this PR.

### Minor / polish (no blockers)
- **Unbounded `_failure_state` + raw-URL keying:** entries are only removed on success, and keying on the raw URL string means any caller putting a changing query string *in the URL itself* (e.g. DRF `next` URLs in `http_request_paginated`) defeats both the throttle (new key → fresh 60 s window → never warns) and the cleanup. A one-line comment documenting "keys must be stable across polls" would make the constraint explicit.
- **Untested kwargs path:** `make_retry_state` supports `args_has_url=False` (kwargs branch of the URL extraction) but all four tests use the default — add one test or drop the parameter.
- **`cast(str, …)` at http.py:39 can evaluate to None** if signature assumptions shift; an `assert` or fallback would be more honest (no risk today — `_before_sleep` is module-private, single decorator).
- **URL now appears in the WARNING message** (`Still failing to reach %s…`), which `before_sleep_log` never logged. Fine for portal URLs (no secrets), worth a sentence in the PR body.
- **Nits:** `__wrapped__` in `test_success_clears_failure_state` bypasses the decorator wiring (reasonable, just noting); `_failure_state` is lock-free (safe under one asyncio loop, fine as-is).

**Bottom line:** correct, well-tested, matches the issue spec. Merge after a quick call on items 1 and 2.

thusser:
Pushed a follow-up commit addressing both decisions:

**1. Caller-level ERROR spam** — extracted the throttle logic into a reusable `LogThrottle` class (`pyobs/utils/http.py`). `PortalTaskArchive`/`PortalObservationArchive` now each own a `LogThrottle(quiet_for=0.0, interval=60.0)`: first poll failure still logs at ERROR immediately (you want to know right away), repeats during the same outage are throttled to at most once a minute instead of once per ~20s cycle, and it clears on the next successful poll.

**2. Changelog version/date** — confirmed intentional: `vX.Y.Z.devN` entries land without bumping `pyproject.toml` (the bump happens at actual release commits, e.g. `38f0851f v2.1.0`), matching the pattern already used for `v2.0.1.dev1` etc. No change needed.

**Minor/polish**, all addressed:
- `LogThrottle`'s docstring now states the URL-key-stability requirement explicitly.
- Untested kwargs-extraction branch in `make_retry_state` now has a dedicated test (`test_before_sleep_extracts_url_from_kwargs`).
- `cast(str, ...)` replaced with `assert isinstance(url, str), ...` so a future signature change fails loudly instead of silently coercing `None`.
- Left the URL-in-WARNING-text point as-is (no secrets in portal URLs, and it's genuinely useful context for the reader).

Also re-ran the full `tests/robotic` suite (392 passed) since the archive files changed.

Issue #832, opened 2026-08-31
Dark masters per exposure time: match science frames by exptime, scale only a reference (600 s) dark


thusser:
## Summary

Reduction side of taking darks at the night's science exposure times (robotic side: #831). Today the pipeline does not care about dark exposure times:

- `Reduction._create_master_calib()` (`pyobs/utils/pipeline/reduction.py`) combines **all** raw darks of a night into a **single** master dark per instrument/binning, regardless of exposure time. The master filename pattern `FILENAME` has no exposure-time component, so per-exposure-time masters would collide on the same name.
- `ReductionBase._master_frames` cache key is `(ImageType, instrument, binning, filter_name)` - no exposure time.
- `Pipeline.find_master()` (`pyobs/utils/pipeline/pipeline.py`) matches masters only by type/instrument/binning/filter/time window (nearest `DATE-OBS`), never by exposure time.
- `Calibration` (`pyobs/images/processors/calibration/calibration.py`) then **always** rescales the dark master linearly to the science exposure time (`dark_scale=True` in `_CCDDataCalibrator` via `ccdproc.ccd_process`).

Goal: **only scale darks for one reference exposure time (configurable, default 600 s); for every other exposure time an individual master dark must exist and be used unscaled.**

## Proposed changes

1. **Group raw darks by exposure time when creating masters** (`Reduction._create_master_calib`, `ImageType.DARK`)
   - List the night's raw darks per instrument/binning **including `EXPTIME`** (uses the archive changes from #831).
   - Group by exposure time within a tolerance (suggest relative 1% or absolute 0.1 s, configurable).
   - Create one master dark per group (bias-subtracted via `create_master_dark`); keep the existing minimum-frames check (>= 3) per group and log/skip groups with too few frames.
   - Extend the `ReductionBase._master_frames` cache key to `(ImageType, instrument, binning, filter_name, exptime)` for darks (other types unchanged / `None`).
   - Make sure the master's `EXPTIME` header survives `ccdproc.combine`; set it explicitly after combining if not.

2. **Distinct filenames for per-exposure-time masters**
   - Add `EXPTIME` to the default `FILENAME` pattern in `pyobs/utils/pipeline/reduction.py` (e.g. `{EXPTIME|...}`); add a formatting function to `FilenameFormatter` (`pyobs/utils/fits.py`) if a clean rendering (600, not 600.0) is wanted.
   - Keep the old pattern valid for existing deployments or document the rename.
[...]

Issue #831, opened 2026-08-31
Take morning darks at the exposure times used for science frames during the night


thusser:
## Summary

Morning calibration darks are currently taken with one fixed exposure time (`DarkBiasScript.exptime`, typically 600 s), regardless of the exposure times actually used for science frames during the night. At calibration time the dark master is then linearly rescaled to the science exposure time (`dark_scale=True` in `ccdproc.ccd_process`, see `pyobs/images/processors/calibration/_ccddata_calibrator.py`). Linear dark scaling is only valid in the linear regime of the dark current and is a known source of error/noise for long exposures or non-linear detectors.

This issue covers the **robotic/observing half**: pyobs-core must be able to determine which exposure times were used for science (OBJECT) frames during the previous night, and the morning calibration must be able to take darks with exactly those exposure times. The reduction half (per-exposure-time masters, exposure-time matching, scaling policy) is tracked in #@@CORE2@@.

## Current state

- Science exposure times are chosen at runtime per target/filter/binning by `pyobs.robotic.utils.exptime.ExposureTimeProvider` (stellar scaling) and the imaging scripts -> a night typically contains several different exposure times.
- Morning darks are taken by `pyobs.robotic.scripts.calibration.DarkBiasScript`: a single `exptime` field (0 -> bias), `count` exposures, one `IMAGETYP` per run.
- The archive API carries **no exposure-time information**:
  - `Archive.list_frames()` / `Archive.list_options()` (`pyobs/robotic/utils/archive/archive.py`) have no `exptime` parameter.
  - `FrameInfo` has only id/filename/filter_name/binning/dateobs - no `EXPTIME`.
  - `LocalArchive._update_root()` already parses headers into a DataFrame but does not include `EXPTIME`; `PyobsArchive` neither queries nor returns it.

## Proposed changes

1. **Expose exposure time in the archive API**
   - Add `exptime: float | None = None` to `FrameInfo`.
   - Add an optional `exptime: float | None = None` filter to `Archive.list_frames()` and `Archive.list_options()` (abstract + both implementations `PyobsArchive`, `LocalArchive`).
   - `PyobsArchive`: include `EXPTIME` in `PyobsArchiveFrameInfoDict`/`PyobsArchiveFrameInfo` and map it in `_build_query()`.
   - `LocalArchive`: read `EXPTIME` in `_update_root()` and filter in `_filter_data()`.
[...]

PR #840, opened 2026-09-01
Take morning darks at the night's science exposure times (#831)

thusser:
## Summary
- Exposes `EXPTIME` on the archive API: `FrameInfo.exptime`, an `exptime` filter/tolerance param on `list_frames()`/`list_options()`, and an `exptimes` key in `list_options()`'s return, in both `PyobsArchive` and `LocalArchive`.
- Adds `pyobs.robotic.utils.calibration.science_exptimes_for_night()`, deriving the distinct (tolerance-grouped) exptimes a night's science frames used, per instrument/binning. Caches its result at module scope for 5 minutes, keyed by `(site, night, tolerance, min_exptime)`.
- Adds `pyobs.utils.exptime_grouping` (`exptimes_close`/`group_exptimes`), a shared relative-tolerance helper reused by the archive filter and the grouping helper (and intended for #832's per-exptime dark-master grouping later).
- Extends `DarkBiasScript` with `exptimes` (explicit list) and `match_science_exptimes` (derived), running one dark series per exptime, longest-first; the single-`exptime` default path (including bias, `exptime=0`) is unchanged. Mutual exclusivity between `exptime`/`exptimes`/`match_science_exptimes` is validated at construction.
- Exptimes below `dark_min_exptime` (default 5s, per ADR 0015) are dropped before grouping, since calibration treats them as bias-only and never needs a dark master for them.
- ADR `0015-dark-master-strict-exptime-matching-reference-scale-down-only.md` flipped from `proposed` to `accepted`; plan `2026-09-01-morning-darks-match-science-exptimes.md` flipped to `implemented`.

Scoped to issue #831 only (archive/robotic side). #832 (reduction/`Calibration` matching+scaling policy) is a separate follow-up plan, not touched here.

## `estimate_duration()` vs. `match_science_exptimes`
`Task.create_script()` re-validates a fresh `Script` (and `Archive`) instance on every `can_run()`/`estimate_duration()` call, so instance-level caching between them doesn't work. Instead, `science_exptimes_for_night()`'s cache lives at module scope: `can_run()` (async) warms it, and `estimate_duration()` (sync, can't query the archive itself) reads it via a new synchronous `peek_cached_science_exptimes_for_night()`. Falls back to a single-series placeholder estimate if nothing is cached yet (e.g. `can_run()` hasn't run for that site/night within the cache's 5-minute TTL). `can_run()` now also surfaces an unreachable archive as a cant-run reason instead of letting that failure surface later during `run()`.

## Other known gaps (called out in the plan's acceptance criteria, not solved by this PR)
- `PyobsArchive.list_options()`'s `exptimes` key depends on the pyobs-archive server actually returning one — not verifiable from this repo. `list_frames()` is safe either way: it re-filters by tolerance client-side regardless of server-side `EXPTIME` support.
- No CHANGELOG entry: the top-of-file version header (`v2.1.1.dev1`) is already stale relative to the just-released `v2.1.1` tag, a pre-existing gap unrelated to this PR — didn't want to guess a version number and add more noise there.

## Test plan
- [x] `pytest tests/` (excluding `integration`/`xmpp`): 1718 passed, 7 pre-existing/unrelated failures in `tests/cli/test_pyobsd.py` (local-environment `pyobs` executable lookup, present on `develop` too).
- [x] `ruff check` clean.
- [x] `pyrefly check` (CI's actual invocation, `pyobs/` only): 0 errors, same baseline suppressed/warning counts as `develop`.
- New tests: `tests/utils/test_exptime_grouping.py`, `tests/robotic/utils/test_calibration.py` (including cache hit/expiry/clear coverage), plus exptime coverage added to `tests/robotic/utils/archive/test_{local,pyobs}_archive.py` and `tests/robotic/scripts/test_darkbias.py` (including the cache-warm-up/fallback split for `estimate_duration()`).

### DIFF
--- a/pyobs/robotic/scripts/calibration/darkbias.py
+++ b/pyobs/robotic/scripts/calibration/darkbias.py
@@ -185,18 +185,22 @@ class DarkBiasScript(Script):
             if self.archive is None or self.site is None:
                 self._cant_run_reason = "match_science_exptimes requires archive and site to be configured."
                 return False
-            night = self._resolve_night()
-            if night is None:
-                self._cant_run_reason = "No observer configured to derive the night from."
-                return False
-            # Task.create_script() re-validates a fresh Script (and Archive) on every call, so
-            # estimate_duration() can't reuse this instance's state later -- warm the module-level
-            # cache here instead, so its later (sync, archive-less) lookup can hit it.
+            # _resolve_night() can raise (e.g. night_obs()'s IERS lookup failing) -- guard it
+            # alongside the archive query so any failure here surfaces as "cannot run" rather
+            # than propagating out of can_run() as a hard exception.
             try:
+                night = self._resolve_night()
+                if night is None:
+                    self._cant_run_reason = "No observer configured to derive the night from."
+                    return False
+                # Task.create_script() re-validates a fresh Script (and Archive) on every call, so
+                # estimate_duration() can't reuse this instance's state later -- warm the
+                # module-level cache here instead, so its later (sync, archive-less) lookup can
+                # hit it.
                 await science_exptimes_for_night(self.archive, self.site, night)
             except Exception:
-                log.exception("Could not query archive for science exptimes.")
-                self._cant_run_reason = "Could not query archive for science exptimes."
+                log.exception("Could not determine night or query archive for science exptimes.")
+                self._cant_run_reason = "Could not determine night or query archive for science exptimes."
                 return False
 
         # seems alright
@@ -285,10 +289,10 @@ class DarkBiasScript(Script):
             return sum(self.count * (e + readout) for e in self.exptimes)
 
         if self.match_science_exptimes:
-            if self.site is not None:
+            if self.site is not None and self.archive is not None:
                 night = self._resolve_night()
                 if night is not None:
-                    cached = peek_cached_science_exptimes_for_night(self.site, night)
+                    cached = peek_cached_science_exptimes_for_night(self.archive, self.site, night)
                     if cached is not None:
                         exptimes = self._flatten_matching_exptimes(cached)
                         if exptimes:
--- a/pyobs/robotic/scripts/calibration/darkbias.py
+++ b/pyobs/robotic/scripts/calibration/darkbias.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Annotated
+from typing import TYPE_CHECKING, Annotated, Self
 
-from pydantic import Field
+from pydantic import Field, model_validator
 
 from pyobs.interfaces import (
     IBinning,
@@ -13,11 +13,13 @@ from pyobs.interfaces import (
     IWindow,
 )
 from pyobs.robotic.scripts import Script
+from pyobs.robotic.utils.archive import Archive
+from pyobs.robotic.utils.calibration import science_exptimes_for_night
 from pyobs.utils.enums import ImageType
+from pyobs.utils.time import Time
 
 if TYPE_CHECKING:
     from pyobs.robotic.task import TaskData
-    from pyobs.utils.time import Time
 
 log = logging.getLogger(__name__)
 
@@ -30,7 +32,35 @@ class DarkBiasScript(Script):
     )
     count: int = Field(default=20, description="Number of exposures to take.")
     exptime: float = Field(default=0, description="Exposure time in seconds. 0 takes a bias, anything else a dark.")
+    exptimes: list[float] | None = Field(
+        default=None, description="Explicit list of dark exposure times to take one series each of."
+    )
+    match_science_exptimes: bool = Field(
+        default=False,
+        description="Derive the dark exposure times from the night's science frames instead of a fixed list "
+        "(requires archive and site). Mutually exclusive with exptime/exptimes.",
+    )
+    archive: Archive | None = Field(
+        default=None, description="Archive to query for science exptimes. Required when match_science_exptimes=True."
+    )
+    site: str | None = Field(
+        default=None, description="Site code to query the archive for. Required when match_science_exptimes=True."
+    )
+    night: str | None = Field(
+        default=None,
+        description="Night to derive science exptimes for (Archive.list_frames(night=...) format). Defaults to "
+        "the night that just ended, derived from the current time.",
+    )
     binning: tuple[int, int] = Field(default=(1, 1), description="Detector binning as (x, y).")
+
+    @model_validator(mode="after")
+    def _validate_exptime_mode(self) -> Self:
+        modes_set = sum([self.exptimes is not None, self.match_science_exptimes])
+        if modes_set > 1:
+            raise ValueError("exptimes and match_science_exptimes are mutually exclusive.")
+        if modes_set == 1 and self.exptime != 0:
+            raise ValueError("exptime cannot be combined with exptimes or match_science_exptimes.")
+        return self
 
     async def can_run(self, data: TaskData | None) -> bool:
         """Whether this config can currently run.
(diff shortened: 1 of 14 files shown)


thusser:
## Review

Nice, well-scoped PR — the archive/exptime plumbing, the shared tolerance helper, `science_exptimes_for_night()`, and the multi-series `DarkBiasScript` all match the plan and are well tested. I verified: `pytest` on all touched test files (84 passed), `ruff` clean, `pyrefly check pyobs/` → 0 errors.

A few things to consider before merge:

**1. (Medium) `DarkBiasScript` unions away the per-combo keying `science_exptimes_for_night` produces.**
`_resolve_exptimes` collapses all `(instrument, binning)` groups into one global set. On a multi-instrument site, camera A would take darks at exptimes only camera B used — at A's fixed configured binning, no less, so some series can't correspond to any science frame A ever took. The plan even says to log the resolved list *per instrument/binning* (section 3), which this doesn't do. The script knows its own `binning`, so it could at least filter by that. If the union is intentional (conservative over-approximation), it should be documented as such.

**2. (Medium) Edge semantics for empty/zero exptimes.**
- With `match_science_exptimes` on a night with no qualifying science frames, `run()` completes having exposed nothing — but only *after* setting the camera to `DARK` image type (`ImageType.BIAS if exptimes == [0] else ImageType.DARK` turns `[]` into DARK). Silent no-op with camera state left modified; consider failing loudly or taking bias.
- The validator accepts `exptimes=[0, 30.0]`, which runs a 0 s exposure under image type DARK — a "dark" that should really be a bias series — while the plan states bias stays its own single series. Reject `0` inside `exptimes` or set image type per series.

**3. (Low/Med) `estimate_duration` for `match_science_exptimes` falls back to `exptime=0` → `count * 5s`.**
Documented as a known gap, and I agree a sync method can't run the archive query — but the scheduler plans around this estimate, and ~100 s for a real ~2 h (20×600 s) run is a severe underestimate. A conservative placeholder (e.g. the ADR's 600 s reference) would be safer than 0.

**4. (Low) `exptimes_close` is asymmetric at the tolerance boundary.** `exptimes_close(100, 101.01)` is `True`, `exptimes_close(101.01, 100)` is `False` (tolerance is relative to the second argument). The archive filter always calls it in a consistent direction, so no practical bug, but #832 will reuse this — worth documenting "tolerance relative to `b`" or making it symmetric.

**5. (Low) `group_exptimes` is a greedy median-drift heuristic, not transitive closure.** E.g. `group_exptimes([100.0, 100.9, 101.8], 0.01)` → `[100.45, 101.8]` even though 100.9 and 101.8 are pairwise within 1%. The docstring overstates the guarantee; results can shift as values are added. Probably fine for discrete real-world exptimes, but #832's reuse may want union-find grouping.

**6. (Low) The `PyobsArchive` client-side re-filter comment overstates its guarantee.** It correctly handles a server that *ignores* EXPTIME (the tested case), but if the server exact-filters on `EXPTIME`, near-matches are dropped before the client sees them and the tolerance re-filter has nothing to recover. Harmless for the main path (the helper never passes `exptime` to `list_frames`), but "an unsupported or exact-only param doesn't matter" is only true in one direction.

**7. (Nit) `PyobsArchiveFrameInfoDict` declares `EXPTIME: float` as required while the code uses `self.info.get("EXPTIME")`** — should be `NotRequired[float]`.

**8. (Process) Plan section 4's deliverable** — documenting the new `DarkBiasScript` fields in `docs/source` — is unimplemented even though the plan is marked *implemented* (the acceptance criteria don't cover it). Also, ADR 0015 flips to *accepted* while the `dark_scale_*` policy it records is #832's, which isn't implemented — defensible since this PR consumes `dark_min_exptime`, but worth an explicit note.

Nits: `sum([bool, bool])` in the validator is opaque; the validator messages name only `exptimes`/`match_science_exptimes` even though `exptime` is the third mutually-exclusive mode.

**Verdict:** approve with minor feedback. Items 1 and 2 I'd want resolved (or explicitly documented as intended) before merge; the rest can be follow-ups.

thusser:
Addressed in 411a2dc7:

1. **Union across instrument/binning** — fixed. `_flatten_matching_exptimes()` now narrows to this script's own `binning` before unioning (instrument still can't be filtered — no INSTRUME-equivalent config field on `DarkBiasScript` — documented as a known limitation in the docstring). New test: `test_only_uses_exptimes_from_its_own_binning`.
2. **Empty/zero exptime semantics** — fixed both: `run()` now returns before touching `IImageType`/`IExposureTime` when `_resolve_exptimes()` comes back empty (camera left untouched, not silently switched to DARK); `exptimes` now rejects `0` at validation (bias stays its own single series, per the plan).
3. **`estimate_duration` fallback** — now uses `_FALLBACK_MATCH_EXPTIME = 600.0` (ADR 0015's reference exptime) instead of the always-0 configured `exptime` when nothing's cached yet.
4. **`exptimes_close` asymmetry** — fixed, now symmetric (`tolerance * max(abs(a), abs(b))`). New test: `test_exptimes_close_is_symmetric`.
5. **`group_exptimes` transitivity** — docstring now states the greedy, non-transitive behavior explicitly rather than implying a clustering guarantee. Left the algorithm as-is per your own "probably fine for discrete real-world exptimes" call — happy to swap to union-find if #832 needs it.

Full suite still green (1721 passed, same 7 pre-existing/unrelated `test_pyobsd.py` failures), `ruff`/`pyrefly` clean.

thusser:
## Re-review (after `e49a955b` + `411a2dc7`)

Thanks for the quick follow-up — I re-verified the new head (`411a2dc7`): `ruff` clean, `pyrefly check pyobs/` → 0 errors, touched test files 95 passed / 1 failed (failure is environmental, see A).

### Addressed ✅

- **Per-combo union (finding 1):** `_flatten_matching_exptimes()` now narrows to the script's own binning and unions across instruments only at that binning, with an honest docstring on why instrument filtering isn't possible (no INSTRUME config field on the script). Warning now names the binning. Covered by `test_only_uses_exptimes_from_its_own_binning`.
- **Empty/zero exptimes (finding 2):** `run()` early-returns before touching image type/exptime when nothing resolves (`set_image_type` not called — asserted), and the validator rejects `0 in exptimes` with a clear "bias is always its own single series" message.
- **`estimate_duration` (finding 3):** nice solution — 300 s TTL module-level cache in `calibration.py` warmed by `can_run()`'s async query, read by the sync `estimate_duration()` via `peek_cached_science_exptimes_for_night()`, cold fallback to `_FALLBACK_MATCH_EXPTIME = 600.0` (ADR 0015's reference) instead of the always-0 `exptime`. Tests cover the warm handoff (fresh instance), cold fallback, TTL expiry, and `clear_cache()`.
- **`exptimes_close` asymmetry (finding 4):** now `tol * max(abs(a), abs(b))`, documented symmetric, tested.
- **`group_exptimes` (finding 5):** docstring now honestly documents the greedy median-drift behavior with the exact `[100.0, 100.9, 101.8]` example.

### Still open (minor, from before)

- PyobsArchive's "an unsupported or exact-only param doesn't matter" comment still overstates: with an exact-only server, near-matches are dropped before the client re-filter can recover them. Harmless for the main path (the helper never passes `exptime` to `list_frames`).
- `PyobsArchiveFrameInfoDict.EXPTIME` is typed required but read via `.get()` — should be `NotRequired[float]`.
- Plan §4's docs deliverable (document the new script fields in `docs/source`) is still unimplemented while the plan is marked *implemented*.

### New on the follow-up

**A. (Should fix before merge) `test_can_run_true_with_observer_configured` now needs network.** It uses a real `astroplan.Observer` and `Time.now().night_obs()` → astropy IERS auto-download. Offline (or with a read-only astropy cache, or a clock beyond the bundled IERS coverage — all three here) it fails with `ValueError` from `iers.py`. Every other Observer-based test in the repo uses fixed in-coverage times, so the suite previously ran fully standalone. It'll probably pass in CI (network), but consider pinning IERS (`conf.auto_max_age = None` + bundled file) or mocking `Time.now()` to keep the suite standalone.

**B. (Design note) `can_run()` is no longer cheap/local for `match_science_exptimes`.** It now executes the full archive query (list_options + per-combo, paginated list_frames for PyobsArchive) to warm the cache — bounded to once per (site, night) per TTL, but the first call after expiry adds real latency to a scheduling-tick method, and an archive outage now reports "cannot run" instead of failing at run time. Deliberate and documented; fine. One robustness gap: `_resolve_night()` sits *outside* the `try/except`, so an IERS failure in `night_obs` propagates out of `can_run()` as a hard exception rather than a cannot-run reason — that's exactly the failure mode test A hits here. Consider moving it inside the guarded block.

**C. (Nit) Cache key omits archive identity** — `(site, night, tolerance, min_exptime)` only; two archives serving the same site/night in one process would collide for up to 5 min. Acknowledged in the code comment; unlikely in practice.

**D. (Nit) Estimate depends on call order** — real numbers only if `can_run()` ran first within the TTL; cold calls get the 600 s placeholder (sensible, and tested).

### Verdict

All substantive findings resolved with good tests. Ship after A (make the observer test IERS-independent); B–D and the three "still open" items can be follow-ups.

Issue #847, opened 2026-09-01
Scheduler never reschedules when a task is deactivated in the portal (stale observations keep mastermind in 'Could not resolve task' loop)


thusser:
## Symptom

When a task is deactivated (`active=false`) or deleted in pyobs-portal while it still has pending observations, the mastermind gets stuck in an error loop and the scheduler never reschedules:

```
21:48:28 INFO  observationarchive.py:110 Downloaded new schedule. Next observation is task wasp45b at 2026-08-31T21:53:23.206.
21:53:24 ERROR observationarchive.py:182 Could not resolve task for observation 149153, skipping.
21:53:34 ERROR observationarchive.py:182 Could not resolve task for observation 149153, skipping.
... (repeats every ~10 s until the observation window passes)
```

The window freed by the deactivated task is never refilled and the stale observation stays pending in the portal.

## Root cause

1. pyobs-portal's `/api/tasks/` filters `active=True` (`pyobs_portal/api/views.py`), so a deactivated task simply disappears from the task list.
2. `PortalTaskArchive` (pyobs-core) detects the removal on its next poll (unconditional refetch since `specs/plans/2026-08-20-backend-archive-marker-loss.md`) and fires `on_tasks_changed` → `Scheduler._update_schedule()`.
3. `_update_schedule()` (`pyobs/modules/robotic/scheduler.py:213-218`) gates rescheduling on "was the removed task actually scheduled?", answered via `self._schedule.get_schedule()`. For `PortalObservationArchive` that returns its **cache**, which the Scheduler deliberately leaves empty (`auto_update=False`, injected since commit 422bfcd6 "Fix Scheduler kwarg injection: gate auto_update on target class support"). So `removed_from_schedule` is always `[]` → "Found N tasks, but none of them was scheduled." → `_need_update = False` → **no reschedule**.
4. Deactivation does not cascade to observations, so the mastermind's `PortalObservationArchive` never sees observation 149153 change. `fetch_task()` fails (`PortalTaskArchive.get_task` returns `None` for the deactivated task) → "Could not resolve task for observation 149153, skipping." every mastermind loop iteration (~10 s).

Net effect: the task-change signal is received but discarded by the scheduler gate; the stale observation is never removed from the pending set.

## Proposed fix

[...]

PR #852, opened 2026-09-01
Reschedule on portal task removal, self-heal unresolvable observations

thusser:
## Summary

Fixes #847: when a task is deactivated or deleted in pyobs-portal while it has pending
observations, `Scheduler._update_schedule()` never reschedules and the mastermind gets stuck
logging `Could not resolve task for observation N, skipping.` every ~10s until the observation's
window passes on its own.

- **Scheduler**: drop the "was the removed task actually in schedule?" gate
  (`_update_schedule()`). For `PortalObservationArchive` the consulted cache
  (`get_schedule()` → `self._observations`) is permanently empty by construction (the Scheduler
  runs it with `auto_update=False` so its own poller doesn't double-drive updates), so the gate
  silently discarded every portal task removal. It was a 2022 optimization (`b3464f89`), never a
  correctness requirement — dropping it also removes a latent `AttributeError` (`s.task.id` on a
  bare-int FK, same shape guarded in `lco/observationarchive.py:99`).
- **PortalObservationArchive**: `get_next_observation()`/`get_current_observation()` now mark an
  observation `CANCELED` (locally and on the portal) when `fetch_task()` can't resolve its task,
  instead of logging-and-skipping it forever. A failed cancel PUT is caught and logged, not
  propagated.

Plan: `specs/plans/2026-09-01-scheduler-reschedule-on-portal-task-removal.md`

## Test plan

- [x] `pytest tests/modules/robotic/ tests/robotic/` — 516 passed
- [x] `ruff check` / `black --check` on touched files
- [x] `pyrefly check` on touched files

https://claude.ai/code/session_01AUQPZXJSVJjUiZjKNT6yhY

### DIFF
--- a/pyobs/modules/robotic/scheduler.py
+++ b/pyobs/modules/robotic/scheduler.py
@@ -209,14 +209,6 @@ class Scheduler(Module, IRunnable, IRoboticScheduler):
             log.info("Only one removed block detected, which is the one currently running.")
             self._need_update = False
 
-        # check, if one of the removed blocks was actually in schedule
-        if len(removed) > 0 and self._need_update:
-            schedule = await self._schedule.get_schedule()
-            removed_from_schedule = [s for s in schedule if s.task.id in removed]
-            if len(removed_from_schedule) == 0:
-                log.info("Found %s tasks, but none of them was scheduled.", len(removed))
-                self._need_update = False
-
         # store blocks
         self._tasks = tasks
 
--- a/pyobs/robotic/storage/portal/observationarchive.py
+++ b/pyobs/robotic/storage/portal/observationarchive.py
@@ -37,6 +37,9 @@ class PortalObservationArchive(ObservationArchive):
         # First failure logs immediately at ERROR (someone should know right away); repeats
         # during the same outage are throttled to at most one ERROR per minute.
         self._poll_error_throttle = LogThrottle(quiet_for=0.0, interval=60.0)
+        # Same pattern, keyed per observation id: a portal outage would otherwise re-log an
+        # ERROR every poll for as long as the cancel PUT keeps failing.
+        self._cancel_error_throttle = LogThrottle(quiet_for=0.0, interval=60.0)
 
         if auto_update:
             self.add_background_task(self._check_for_changes)
@@ -171,22 +174,37 @@ class PortalObservationArchive(ObservationArchive):
         """
         return self._observations
 
-    async def _resolve_or_cancel(self, obs: Observation, task_archive: TaskArchive) -> bool:
-        """Resolve obs.task; if the portal no longer has the task, mark obs canceled so it stops
-        being retried on every subsequent poll instead of logged-and-skipped forever.
+    async def _resolve_or_cancel_pending(self, obs: Observation, task_archive: TaskArchive) -> bool:
+        """Resolve a *pending* obs.task; if the task archive has completed at least one poll and
+        the portal no longer has the task, mark obs canceled so it stops being retried on every
+        subsequent poll instead of logged-and-skipped forever.
+
+        Only for pending observations: an in-progress one should be allowed to finish even if its
+        task was deactivated mid-run, not flipped to canceled out from under the mastermind.
 
         Returns:
             True if obs.task resolved and obs is usable.
         """
+        task_id = obs.task
         await obs.fetch_task(task_archive)
         if obs.task is not None:
             return True
+        if await task_archive.last_changed() is None:
+            # task archive has never completed a poll (e.g. still starting up) -- a genuinely
+            # missing task looks the same as this transient race, so don't cancel on it.
+            log.error("Could not resolve task for observation %s, skipping.", obs.id)
+            return False
         log.error("Could not resolve task for observation %s, marking canceled.", obs.id)
+        obs.task = task_id  # portal's task FK is non-nullable -- keep the id for the PUT payload
         obs.state = ObservationState.CANCELED
         try:
             await self.update_observation(obs)
+            self._cancel_error_throttle.clear(str(obs.id))
         except Exception as e:
-            log.error("Failed to mark observation %s canceled on portal: %s", obs.id, e)
+            if self._cancel_error_throttle.should_escalate(str(obs.id)):
+                log.error("Failed to mark observation %s canceled on portal: %s", obs.id, e)
+            else:
+                log.debug("Failed to mark observation %s canceled on portal: %s", obs.id, e)
         return False
 
     async def get_next_observation(self, time: Time, task_archive: TaskArchive | None = None) -> Observation | None:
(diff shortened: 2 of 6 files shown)


thusser (commented):
## Review

Reviewed against the plan in `specs/plans/2026-09-01-scheduler-reschedule-on-portal-task-removal.md` (#847), the scheduler/mastermind call paths, the portal API, and the tests. All 93 tests in the touched files pass; ruff is clean. (5 `test_yaml_archives.py` failures are pre-existing astropy IERS table issues on a read-only cache, unrelated.)

**Part 1 — dropping the "was it scheduled?" gate in `Scheduler._update_schedule()`: correct, this fixes the reported bug.** The diagnosis checks out: `PortalObservationArchive.get_schedule()` returns `_observations`, permanently empty when the Scheduler drives the archive with `auto_update=False`, so the gate always suppressed rescheduling. The latent `AttributeError` claim (`s.task.id` on a bare-int FK) is accurate — same shape LCO guards at `lco/observationarchive.py:99`. No feedback loop either: `_update_schedule` fires only from the task archive's `on_tasks_changed`. Note that the actual cleanup of orphaned observations in the scheduler setup happens via the reschedule's `clear_schedule()` (`/api/cancel_observations/`), which cancels all pending obs with `end >= start` — that's what ends the stuck mastermind loop.

**Part 2 — `_resolve_or_cancel()`: the cancel never lands on the portal (must fix).** After a failed `fetch_task()`, `obs.task` is `None`, and `update_observation()` → `model_dump(use_task_id=True)` sends `task: null`. The portal's `Observation.task` FK is non-nullable (`pyobs_portal/api/models.py:78`), and DRF rejects the payload — verified against the portal serializer: `errors: {'task': [ErrorDetail(string='This field may not be null.', code='null')]}`. The exception is caught/logged as designed, but the cancel **never persists**: every poll re-fetches the obs as `pending`, re-flips it to canceled locally, re-PUT-fails — a permanent loop that's now noisier than before (two ERROR lines per cycle).

This also invalidates the plan's escape hatch ("the observation will also disappear from the next `_get_schedule()` fetch regardless"): the portal observation filter (`pyobs_portal/api/filters.py`) filters by state/time only, not task activity, and deactivation doesn't cascade (FK cascade fires only on hard delete).

Suggested fix — preserve the FK id for the cancel payload:

```python
task_id = obs.task
await obs.fetch_task(task_archive)
if obs.task is not None:
    return True
obs.task = task_id          # portal's task FK is non-nullable
obs.state = ObservationState.CANCELED
try:
    await self.update_observation(obs)
...
```

**Design concern: one-shot cancel can misfire on a stale/empty task cache.** `PortalTaskArchive.get_task()` returns `None` for "not in the last successfully-polled list" — which includes the *never-populated* window (boot, archive restart, first poll failing). `None` is a truthful signal for network failures (last-known-good list is retained), but not for the initial-empty window: at startup the mastermind's observation and task archives poll concurrently, and if the observation poll lands first, `_tasks` is empty → every pending obs gets canceled. A transient local flip today, but a *permanent wrongful cancellation on the portal* once the payload fix lands. There's a ready-made guard: `PortalTaskArchive.last_changed()` is `None` until the first successful poll — skip canceling while the cache has never been populated.

**Design question: canceling an IN_PROGRESS observation via `get_current_observation()` is dubious** — a task deactivated mid-run should probably be allowed to finish, and the portal record flipping to `canceled` while the mastermind executes it gives two divergent state views. This path is currently unreachable in practice (the scheduler's archive cache is always empty; the mastermind never calls `get_current_observation`), but I'd make the intent explicit: only self-heal *pending* observations, keep log-and-skip for in-progress ones.

**Tests:** the four new portal tests are good in intent but never assert the PUT payload — that's why the `task: null` bug sails through. Add e.g. `call_args[1]["json"]["task"] == 99` (fails today). The two new scheduler tests are near-duplicates; the stronger regression is asserting `scheduler._schedule.get_schedule.assert_not_awaited()` to lock in the gate's removal. Minor: the "marking canceled" ERROR fires every poll cycle while a portal hiccup persists — the archive's existing `LogThrottle` pattern could apply here too.

**Summary:** merge Part 1 as-is; Part 2 needs the payload fix (and ideally the cache-populated guard) before it can honestly be called a self-heal.

thusser:
Fixed all three, in c46aaa57:

1. **Payload bug** — `obs.task` is now saved before `fetch_task()` and restored onto `obs` before the cancel PUT, so the payload carries `task: <id>` instead of `null`. Added a regression test asserting `call_args[1]["json"]["task"] == 99`.
2. **Startup-race false-cancel** — guarded via `task_archive.last_changed() is None`; falls back to the original log-and-skip while the task archive hasn't completed its first poll. New test covers it.
3. **`get_current_observation()`** — reverted to the original inline log-and-skip, no cancel. Only `get_next_observation()`'s pending path self-heals now. New test locks this in.

Also added a per-observation `LogThrottle` around the "failed to mark canceled" log line, matching the existing `_poll_error_throttle` pattern, per your minor note.

Updated the plan doc (`specs/plans/2026-09-01-scheduler-reschedule-on-portal-task-removal.md`) with a "Revision from review" section documenting all three bugs and the fixes, and the index status to **implemented**.

All 516 tests in `tests/modules/robotic/` + `tests/robotic/` pass, ruff/black/pyrefly clean.

thusser (commented):
## Re-review: all findings addressed — approving

Reviewed commit `c46aaa57` (follow-up to `149030bb`). All three substantive findings from the first review are fixed and verified; 93/93 tests in the touched files pass, ruff clean. (The 5 `test_yaml_archives.py` failures remain a pre-existing astropy IERS/cache environment issue, unrelated.)

**1. Cancel payload `task: null` — fixed.** `_resolve_or_cancel_pending()` captures `task_id = obs.task` before `fetch_task()` and restores it onto `obs` before the PUT, so the payload carries the FK id the portal's non-null `Observation.task` requires. The new assertion `json["task"] == 99` locks it in — exactly the gap I'd flagged.

**2. Startup race — fixed, safely.** The `task_archive.last_changed() is None` guard skips canceling until the first successful poll. `last_changed()` is abstract on the base `TaskArchive` and implemented by all backends; portal/memory/LCO return `None` only until the first poll, and none of them hit the network from it — no new failure mode in the mastermind's poll loop. Regression test covers the never-polled window (stays `PENDING`, no PUT).

**3. In-progress cancel — fixed.** `get_current_observation()` is back to inline log-and-skip; only the pending path self-heals. Test asserts the obs stays `IN_PROGRESS` with no request.

Also good: per-observation `LogThrottle` on the cancel-failure log (first ERROR, then ≤1/min, cleared on success), the merged scheduler test asserting `get_schedule.assert_not_awaited()`, and the plan doc updated to `implemented` with the revision history — all consistent with the repo's spec conventions.

Non-blocking nits (no need to change before merge): `FilesystemTaskArchive.last_changed()` always returns `None`, so a portal-archive + filesystem-task pairing would silently never self-heal (degrades to log-and-skip; safe, but the guard comment's "never completed a poll" wording doesn't describe that backend); and the "marking canceled" line itself is unthrottled (comparable noise to pre-fix behavior, and the cancel succeeds on first try in the genuine case).

Part 1 (gate removal) was already correct; Part 2's self-heal is now actually effective, race-guarded, and conservative about in-progress observations. Approving.

Issue #855, opened 2026-09-01
PolymorphicBaseModel's custom model_serializer ignores exclude/include (and other model_dump kwargs)


thusser:
## Symptom

`PolymorphicBaseModel.inject_class_on_serialization()` (`pyobs/utils/serialization.py:44-49`) is declared `@model_serializer(mode="wrap")` but never calls the `handler` it's given — it builds the output dict itself via raw `getattr()` over `type(self).model_fields`:

```python
@model_serializer(mode="wrap")
def inject_class_on_serialization(self, handler: ValidatorFunctionWrapHandler) -> dict[str, Any]:
    result = {field_name: getattr(self, field_name) for field_name in type(self).model_fields}
    result["class"] = f"{self.__module__}.{self.__class__.__name__}"
    return result
```

Because it never delegates to `handler(self)`, every `model_dump()`/`model_dump_json()` call on any `PolymorphicBaseModel` subclass silently ignores `exclude`, `include`, `by_alias`, `exclude_none`, `exclude_unset`, `exclude_defaults`, etc. Verified directly:

```python
from pyobs.robotic.task import Task
t = Task(id=1, name="t1", duration=100, updated_at="2026-01-01T00:00:00Z")
t.model_dump(exclude={"updated_at"})
# {'id': 1, ..., 'updated_at': '2026-01-01T00:00:00Z', 'class': 'pyobs.robotic.task.Task'}
# 'updated_at' is still there -- exclude had no effect.
```

`PolymorphicBaseModel` subclasses include at least `Task`, `Script`, `Constraint`, `Merit`, and `Target` (and their concrete leaf types) -- anywhere a caller passes `exclude=`/`include=` to `model_dump()` on one of these gets silently wrong output instead of an error.

## Where this bit us

pyobs-core PR #854 (issue #848) added `Scheduler._changed_task_ids()`, which needed to compare two `Task` dumps while excluding the `updated_at` timestamp field (a portal-side save marker, not scheduling content). `task.model_dump(exclude={"updated_at"})` compiled and ran without error but silently included `updated_at` anyway, which would have made a no-op re-save look like a "changed" task and force spurious reschedules. Caught by a regression test (`test_changed_task_ids_ignores_updated_at`) before merge; worked around locally in that PR via `Scheduler._content_dump()`, which calls `model_dump()` with no `exclude` and pops the key from the resulting dict afterward instead.

## Why not fixed in place

[...]

PR #857, opened 2026-09-02
fix: honor exclude/include/by_alias/exclude_* on PolymorphicBaseModel

thusser:
## Summary

- `PolymorphicBaseModel.inject_class_on_serialization()` (`Task`/`Script`/`Constraint`/`Merit`/
  `Target`'s `@model_serializer`) hand-built its output dict via raw `getattr()` and never touched
  `handler`, so `model_dump()`/`model_dump_json()` silently ignored `exclude`, `include`,
  `by_alias`, `exclude_none`, `exclude_unset`, and `exclude_defaults` on every subclass.
- Adds `info: SerializationInfo` and applies all of the above for flat (top-level, non-nested)
  field specs — the only pattern any current caller uses.
- A nested spec (e.g. `{"constraints": {0: {"cost"}}}`) now raises `NotImplementedError` instead
  of being silently ignored or partially applied. Note: pydantic-core wraps any exception a
  `@model_serializer` raises into its own `PydanticSerializationError` before it reaches the
  caller, embedding the original type/message in the wrapped text — that's pydantic-core's own
  behavior, not something this fix controls.
- Full nested-spec support (matching pydantic's own include/exclude grammar for sequences,
  `__all__`, partial nested excludes) is intentionally out of scope — see
  `specs/plans/2026-09-02-polymorphic-model-dump-exclude-include.md`.

Fixes #855.

## Test plan

- [x] New `tests/utils/test_serialization.py` (15 tests): flat exclude/include (set and
      dict-with-`True` forms), `by_alias`, `exclude_none`/`exclude_defaults`/`exclude_unset`,
      nested-spec rejection (direct and via a nested `Constraint` field), plain-dump regression,
      and abstract-type-resolution regression (subclass-specific fields still present under a
      `list[Constraint]`-typed field).
- [x] Full non-integration suite: 1817 passed (7 pre-existing `tests/cli/test_pyobsd.py` failures
      confirmed present on `develop` before this change too — local "pyobs executable not found",
      unrelated).
- [x] `ruff`, `black`, `pyrefly` clean.

https://claude.ai/code/session_01MJgXh6sdZeSJQ9epwra7H1

### DIFF
--- a/pyobs/modules/robotic/scheduler.py
+++ b/pyobs/modules/robotic/scheduler.py
@@ -268,18 +268,13 @@ class Scheduler(Module, IRunnable, IRoboticScheduler):
 
     @staticmethod
     def _content_dump(task: Task) -> dict[str, Any]:
-        """``task.model_dump()`` with ``updated_at`` stripped.
+        """``task.model_dump(exclude={"updated_at"})``.
 
-        ``Task`` is a ``PolymorphicBaseModel``, whose ``@model_serializer`` builds its own dict
-        from ``model_fields`` and injects a ``class`` key (``pyobs/utils/serialization.py:44-49``)
-        -- it does not honor ``model_dump(exclude=...)`` at all, so the key has to be dropped from
-        the result afterwards instead. ``updated_at`` is a portal-side save timestamp, not
-        scheduling-relevant content, and a no-op re-save (e.g. an unchanged DRF PATCH) still bumps
-        it -- leaving it in would report every no-op edit as a "changed" task.
+        ``updated_at`` is a portal-side save timestamp, not scheduling-relevant content, and a
+        no-op re-save (e.g. an unchanged DRF PATCH) still bumps it -- leaving it in would report
+        every no-op edit as a "changed" task.
         """
-        dump = task.model_dump()
-        dump.pop("updated_at", None)
-        return dump
+        return task.model_dump(exclude={"updated_at"})
 
     @classmethod
     def _changed_task_ids(cls, tasks1: list[Task], tasks2: list[Task]) -> set[Any]:
--- a/pyobs/robotic/storage/filesystem/observationarchive.py
+++ b/pyobs/robotic/storage/filesystem/observationarchive.py
@@ -248,6 +248,12 @@ class YamlObservationArchive(FileSystemObservationArchive):
             return ObservationList([self.pyobs_model_validate(Observation, obs) for obs in observations])
 
     async def _save_observations_to_file(self, path: str, observations: ObservationList) -> None:
+        # exclude_defaults now also applies to the nested (PolymorphicBaseModel) Task/Constraint/
+        # etc. fields, not just Observation's own -- pyobs-core#855 fixed inject_class_on_
+        # serialization() silently ignoring it. Saved YAML files drop more default-valued task
+        # fields (active, empty constraints/merits/script, priority: 1.0, ...) than before; this
+        # is intentional and round-trips fine (model_validate refills the defaults), but anyone
+        # reading these files raw will notice the shape change.
         data = [obs.model_dump(mode="json", exclude_defaults=True) for obs in observations]
         with open(path, "w") as f:
             yaml.safe_dump(data, f)
--- a/pyobs/robotic/storage/portal/taskarchive.py
+++ b/pyobs/robotic/storage/portal/taskarchive.py
@@ -84,21 +84,6 @@ class PortalTaskArchive(TaskArchive):
             await self._update()
             self._last_marker = last_update
 
-    @staticmethod
-    def _task_content_dump(task: Task) -> dict[str, Any]:
-        """``task.model_dump()`` with ``updated_at`` stripped.
-
-        ``Task`` is a ``PolymorphicBaseModel``, whose ``@model_serializer`` builds its own dict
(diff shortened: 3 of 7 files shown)


thusser:
## Review — no blockers

Reviewed head `dcafcdfa` (1 commit on `a3f7bb8f`; 4 files, +210/−12). Fixes #855.

### Verification (independently rerun)

| Check | Result |
|---|---|
| `pytest tests/utils/test_serialization.py` (15 new tests) | 15 passed |
| Full non-integration suite, PR head vs `develop` (same env, clean copies) | **1810 vs 1795 passed, identical 14 failures** on both (7× `tests/cli/test_pyobsd.py`, 1× `tests/mixins/test_fitsheader.py`, 6× `tests/modules/telescope/test_dummysolartelescope.py` — known env-flaky files, unrelated). PR adds exactly its 15 tests, zero regressions. The "1817 passed / 7 failed" claim is consistent: 1810 + the 7 env-flaky tests passing in the author's environment |
| `ruff check` + `black --check` on changed files | clean |
| `pyrefly check` (PR vs develop copies) | 0 errors on both, identical warnings — no new type errors |
| Semantic parity probe: equivalent plain pydantic model vs polymorphic model, 5 instance shapes × 19 flat-kwarg combos (`exclude` set/dict-True, `include`, `by_alias`, `exclude_none`/`exclude_unset`/`exclude_defaults` and mixes, `include`+`exclude` together, `model_dump_json`) | **all outputs identical** (modulo the injected `"class"` key) |

Extra probes all behaved correctly: dict-form `include={"x": True}`; time-varying `default_factory` (uuid/datetime) is *not* falsely excluded under `exclude_defaults` (matches native); a flat child spec forwarded from a native parent into a polymorphic child (e.g. `Observation.model_dump(exclude={"task": {"constraints"}})`) is now honored, including via `model_dump_json`; a genuinely nested spec raises `PydanticSerializationError` embedding `NotImplementedError`, exactly as documented.

### 1. Side effect not called out in the PR: `YamlObservationArchive` file content changes (benign, intended-in-kind)

`pyobs/robotic/storage/filesystem/observationarchive.py:251` dumps `Observation` (plain model) with `model_dump(mode="json", exclude_defaults=True)`; the nested `task` is a polymorphic `Task`. Before, `exclude_defaults` was silently ignored for the nested `Task`/`Constraint`; after, it applies — so newly saved archive YAML drops default-valued task/constraint fields (`active`, empty `constraints`/`merits`/`script`, `priority: 1.0`, `project`, `static_target`, `updated_at`, `cost`/`target_dependent` defaults). Round-trip is safe (validation refills the defaults; no test breaks), but the file shape changes, which anyone reading those YAMLs raw (tooling, operators) would notice. This is the *fix* working as intended, but the PR description only claims "plain `model_dump()` unchanged" — worth one line in the description so the archive-shape change is a conscious decision.

### 2. Two docstrings become false the moment this lands

`Scheduler._content_dump` (`pyobs/modules/robotic/scheduler.py:269-282`) and `PortalTaskArchive._task_content_dump` (`pyobs/robotic/storage/portal/taskarchive.py:87-100`) both state Task's `model_dump` "does not honor `model_dump(exclude=...)` at all" and cite `serialization.py:44-49`. Post-merge both claims are wrong and the line refs are stale. The plan's §5 deliberately defers this cleanup ("optional, purely cosmetic") — fine, but the docstrings will actively mislead; at minimum refresh them, or (better) collapse both helpers to `model_dump(exclude={"updated_at"})` since it's now equivalent.

### Nits / small test gaps

- No regression test pins the newly-enabled *native-parent → polymorphic-child* flat-spec forwarding (finding 1's mechanism, and the only in-repo caller with dump kwargs reaching a polymorphic child). Cheap insurance given the plan's own regression-paranoia.
- The `exclude_computed_fields` guard is dead code today (no subclass has `@computed_field`), as the plan acknowledges — but note the converse stays silently wrong: a future subclass *with* computed fields would still omit them from a plain `model_dump()` (handler bypass, pre-existing). One sentence in the guard's comment would cover it.
- Tests are behavior-assertions only (no oracle comparison against a plain pydantic model); my parity matrix passed, so coverage is adequate, but the natural next test for any regression would be exactly that kind of equivalence check.

### Bottom line

Correct, well-tested, plan-faithful fix. The wrapped-`PydanticSerializationError` caveat is honestly documented in code, plan, and PR body. No blockers; recommend the two follow-ups (finding 2 docstrings, and flagging the YAML-archive shape change) before or right after merge.

thusser:
Addressed both findings in `53c983e9`:

1. **YAML archive shape change**: added a comment at `YamlObservationArchive._save_observations_to_file()`'s dump call documenting that `exclude_defaults=True` now also applies to nested `Task`/`Constraint` fields (previously silently ignored), and why that's intentional/safe (round-trips via `model_validate`).
2. **Stale docstrings**: collapsed `Scheduler._content_dump()` and `PortalTaskArchive`'s task-dump helper to call `model_dump(exclude={"updated_at"})` directly, removing the pop-after-dump workaround (and `PortalTaskArchive._task_content_dump()` entirely, matching how `Project`'s call site already worked). No more false claims about `exclude=` being ignored.

Nits (parity-check regression test, `exclude_computed_fields` comment) left as-is per your own "cheap insurance"/optional framing — happy to add if you'd rather have them before merge.

PR #860, opened 2026-09-02
ImageWatcher: relative-path preservation and recursive watching

thusser:
## Summary

- Adds `flatten: bool = True` to `ImageWatcher`: when `False`, a non-templated destination keeps a
  file's path relative to `watchpath` instead of collapsing it to just the basename, so a nested
  directory tree can be relocated while preserving structure. Default preserves existing behavior
  for every current user.
- Recursive watching: `_watch_inotify` (via `asyncinotify.RecursiveWatcher`, already available at
  the pinned dependency floor), `_watch_poll`, and `open()`'s initial scan (both via `vfs.find()`)
  now pick up subdirectories, not just the watched root.
- While rewriting `_watch_poll`, fixed two pre-existing bugs in the code being touched:
  `poll_interval` was stored but never actually awaited between iterations (a busy-loop), and a
  stray debug `print()` was left in the new-file branch.

## Known limitation (documented, not fixed here)

Confirmed by direct reproduction: when a *destination* directory tree that doesn't exist yet is
created and something is renamed into it before this process's asyncio loop is scheduled to
consume the resulting `CREATE` event and add a watch, the move is never observed — inotify only
reports events for already-watched parents. This is a genuine kernel/library-level race (the
`asyncinotify` example script documents the same caveat), not something fixable at this layer.
Kept as an `xfail(strict=True)` regression test so it stays visible rather than silently assumed
away. A directory whose parent tree already existed at watcher startup is unaffected (covered by a
passing test). See the plan doc for the full writeup and the suggested consumer-side mitigation
(periodic reconciliation re-scan).

## Test plan

- [x] `uv run pytest tests/modules/image/test_imagewatcher.py -v` — 21 passed, 1 xfailed (the known
      limitation above)
- [x] `uv run pytest tests/ -m "not integration and not xmpp"` — full suite, 1832 passed, 1
      pre-existing unrelated failure (confirmed via `git stash` to fail identically without this
      branch's changes — a sandbox permission error writing to `/opt/pyobs/storage`, unrelated to
      `ImageWatcher`)
- [x] `black --check` / `ruff check` clean
- [x] `pyrefly check pyobs/modules/image/imagewatcher.py` — 0 errors

See `specs/plans/2026-09-02-imagewatcher-relative-path-recursive.md` for the full plan, reasoning,
and status.

https://claude.ai/code/session_01TRXBxNFz1KrzhyKaKtdy8Y

### DIFF
--- a/pyobs/modules/image/imagewatcher.py
+++ b/pyobs/modules/image/imagewatcher.py
@@ -60,13 +60,26 @@ class ImageWatcher(Module):
             wait_time: Time in seconds between adding a file to the list and processing it. Gives a file that is
                 still being written time to finish (relevant for poll mode and the initial scan) and spaces out
                 re-queued files after failed destination copies.
-            pattern: Only watch/process files matching this fnmatch pattern; also applied recursively, i.e. it
-                is matched against the full path, so a pattern like "*.fits" matches files in subdirectories too.
+            pattern: Only watch/process files matching this fnmatch pattern, checked against the full path (so
+                e.g. "*.fits" also matches files found in subdirectories). This is the single point of filtering,
+                applied uniformly in ``add_file`` regardless of watch mode -- inotify events, polling, and the
+                initial scan in ``open()`` all discover candidates first and let ``add_file`` decide, none of
+                them pre-filter by pattern before that.
             flatten: For a non-templated destination (one without ``{placeholder}``s), whether to collapse the
                 file's path to just its basename under that destination (the historical behavior, default) or to
                 preserve its path relative to ``watchpath`` under the destination instead. Files found via
                 recursive watching or polling only keep their subdirectory structure at the destination when this
-                is False.
+                is False. Note this relies on the destination VFS root creating missing parent directories on
+                write (the default for e.g. ``LocalFile``) -- a destination configured to not do that would
+                repeatedly fail and re-queue any file whose relative path needs a parent directory that doesn't
+                exist yet.
+
+        Note:
+            If a directory holding a file that's already queued (added but not yet processed by ``_worker``) is
+            renamed before that happens, the worker's read of the file at its captured (pre-rename) path fails
+            and the file is dropped rather than picked up at its new location -- this is unrelated to, and not
+            covered by, the recursive-watching support for directories renamed after they no longer hold queued
+            files. Avoid renaming directories that may still hold queued/retrying files.
         """
         Module.__init__(self, **kwargs)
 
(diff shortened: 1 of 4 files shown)


thusser (commented):
## Review: ImageWatcher relative-path preservation + recursive watching

**Verdict: approve with minor suggestions.** No blocking issues. I verified the claims independently: `tests/modules/image/test_imagewatcher.py` → **21 passed, 1 xfailed** (ran 3×, stable); `ruff`/`black` clean; `pyrefly check pyobs/modules/image/imagewatcher.py` → 0 errors. I also confirmed `RecursiveWatcher` exists in `asyncinotify` **4.2.1** (checked the sdist, not just the installed 4.4.4), and that all VFS backends create parent directories on write (`mkdir=True` default), so `flatten=False`'s nested destination paths work on every backend.

### Good

- `flatten=True` default preserves existing behavior; all pre-existing tests pass unmodified.
- Fixing the `poll_interval` busy-loop and the stray debug `print()` in the code being touched, with a targeted regression test for the sleep, is a nice catch.
- The known CREATE/rename race is real, reproduced, and honestly kept as an `xfail(strict=True)` canary rather than papered over; the plan doc is thorough.
- Cleanup on shutdown is sound: task cancellation closes the inotify handle via `watch_recursive()`'s `finally`.
- No in-repo subclasses/other callers of `ImageWatcher` are affected.

### Findings (non-blocking)

**1. Pattern semantics now diverge between watch modes** (`imagewatcher.py`). Poll mode / `open()` pre-filter through `vfs.find` (basename match per directory) and then `add_file` filters the *full* path again (where `fnmatch` `*` spans `/`); inotify mode filters only via `add_file`'s full-path check. Concrete divergence: pattern `*2026*` with file `/watch/2026/a.txt` — inotify queues it, poll/`open()` drop it at the `find` stage. The new `pattern` docstring ("matched against the full path") describes inotify mode only. Suggestion: make `add_file` the single filtering point (call `find(watchpath, "*")`), or keep the pre-filter but document basename semantics. Low real-world impact (`*.fits` users unaffected).

**2. Plan-doc inaccuracy on `find` for non-local roots** (`specs/plans/2026-09-02-...md` §2: "only `LocalFile` implements `find()` today — other VFS root types raise"). Not true: `VFSFile.find` is a concrete classmethod (flat `listdir` + fnmatch, `pyobs/vfs/file.py`), inherited by SSH/SMB/SFTP/TempFile. Non-local watchpaths in poll mode/`open()` therefore *silently stay flat* rather than raising. Worth fixing the sentence and considering a log warning when a non-local root gets a non-recursive scan.

**3. Move-after-queue orphans source paths** (uncovered gap). Filenames are captured at event/scan time. If a directory holding a *queued* file is renamed before the worker processes it (the consumer's `move_observation_directory` flow — the test only covers writes *after* the rename), the worker's read of the old path fails and the generic handler drops the entry; the file at its new location is never processed (file-level `MOVED_TO` isn't `CLOSE_WRITE`, so renames into watched dirs are never delivered — pre-existing semantics, unchanged). Fine if dirs are only moved once files are processed, but worth a sentence in the plan/PR body: don't move dirs containing queued/retrying files; reconciliation re-scan remains the safety net.

**4. `asyncinotify` bookkeeping can kill the watcher under churn** (robustness). `watch_recursive()`'s internal `add_watch`/`rm_watch` can `OSError` if a directory disappears between event receipt and handling; that exits `_watch_inotify`, and `BackgroundTask` restarts it — self-healing, but repeated churn risks the "succession of failure" quit guard. A `try/except` around the add/remove is cheap insurance.

**5. Nits.** `flatten=False` + a destination root configured `mkdir=False` → infinite requeue/log-spam loop (defaults are `mkdir=True`, so exotic; a docstring note would do). `test_watch_poll_picks_up_nested_files` terminates via mock `side_effect` exhaustion (implicit `StopAsyncIteration`) — works but reads magically. The `event.path is None` guard is effectively dead (yielded events always intersect `CLOSE_WRITE`). §3 empty-dir cleanup deferral is fine as long as it stays tracked.

None of this blocks merge; items 1–2 are quick cleanups, 3–4 are documentation/awareness.

thusser:
Addressed all 5 findings in f9ccc9a9:

1. **Pattern divergence** — fixed. `_watch_poll`/`open()` now call `find(watchpath, "*")` unfiltered; `add_file`'s full-path `fnmatch` check is the single filter point across all three modes. New test: `test_watch_poll_and_inotify_apply_pattern_consistently`.
2. **`find()` plan-doc inaccuracy** — corrected. `VFSFile.find` is a concrete base method (flat, non-recursive), inherited by SSH/SMB/SFTP/TempFile — they don't raise, they silently stay flat. Fixed the wording in the plan doc.
3. **Move-after-queue gap** — documented as a second known limitation (plan doc + class docstring), not fixed, same reasoning as the existing race: mitigation is "don't move a directory that still holds queued/retrying files" + the periodic-reconciliation-rescan recommendation covers it too.
4. **OSError under directory churn** — fixed. `_watch_inotify`'s watch loop now catches `OSError` and restarts with a fresh `RecursiveWatcher` locally rather than letting the task die. New test: `test_watch_inotify_restarts_on_oserror`.
5. **Nits** — removed the dead `event.path is None` guard; rewrote the poll test to use the same explicit sleep-then-cancel pattern as every other test instead of relying on mock exhaustion; documented the `flatten=False` + non-mkdir-destination edge case in the docstring; §3 cleanup deferral needs no change.

23 passed / 1 xfailed on the module's tests, full suite 1834 passed (same one pre-existing unrelated failure), black/ruff/pyrefly clean.

Issue #858, opened 2026-09-02
DomeCapability.estimate_rotate_time_s() -- mean-distance dome rotate time (descoped from live position)


thusser:
## Context

Follow-up to `specs/plans/2026-09-01-instrument-capability-duration-estimates.md`, which feeds
pyobs-portal instrument capability data (slew rate, readout time, filter-change time, etc.) into
`Script.estimate_duration()`. That plan uses a fixed placeholder slew/rotate *distance* (rate alone
isn't a duration) because there's generally no "current pointing" state at estimate time.

One case is solvable now without new design, though: `OnDemandScheduler`'s **first** placed task in
a given `schedule()` call. At that point the live telescope position is genuinely knowable via one
async `IPointingRaDec`/`IPointingAltAz` proxy call, made once by the caller
(`pyobs/modules/robotic/scheduler.py`, which already owns the comm/proxy access) before invoking
`schedule()`, then threaded through the same way `instrument_capabilities` is threaded through
(`TaskScheduler.schedule()` → `TaskData`) — `estimate_duration()` itself stays sync, so the value has
to be pre-fetched, not queried live from inside it.

Two harder related cases are explicitly **not** in scope for this issue (need their own design
first):
- Every task after the first in the same `schedule()` pass — `OnDemandScheduler`'s greedy recursion
  (`schedule_in_interval` → `schedule_first_in_interval`, `ondemandscheduler.py`) doesn't currently
  track "last scheduled task's target" as state across calls.
- pyobs-portal's script-builder live-edit duration estimate (`schema.py`'s `estimate_duration()`) —
  no schedule context and no live telescope reachable from the Django process at all.

## Proposal

- `pyobs/modules/robotic/scheduler.py` fetches the live telescope RA/Dec (or Alt/Az) once per
  `schedule()` invocation, alongside the existing `instrument_capabilities` fetch.
- Thread it through `TaskScheduler.schedule()` / `TaskData` (same shape as
  `instrument_capabilities`).
- `PointingScript`/`ImagingScript`/`AutoFocusScript`'s `estimate_duration()` use it, when present,
  as the start position for the first task's slew-distance calc instead of the fixed placeholder.
  Falls back to the placeholder whenever it's missing (module unreachable, no telescope configured,
  etc.) — same optional/degrade-to-None convention as the rest of the parent plan.

## Repos

pyobs-core only — no pyobs-portal changes needed for this piece.

thusser:
[...]

PR #867, opened 2026-09-03
feat: consume instrument_capabilities in 4 leaf scripts (§A.8)

thusser:
## Summary
§A.8 of `specs/plans/2026-09-01-instrument-capability-duration-estimates.md` — the last piece of §A. Four scripts now read `data.instrument_capabilities` and use real portal-declared numbers wherever available, falling back to today's flat fudge constants whenever the lookup misses at any level (no `data`, no capabilities, no matching `module_name` row, or the specific field not set on the matched row):

- **`ImagingScript`**: adds real per-binning readout time (previously not counted at all) and real per-wheel filter-change time (also previously absent — now counted once per actual filter transition across the full `instrument_configs * repeats` sequence; a config with no `optical_filter` set never counts as a transition, since there's no way to know whether that's actually a change). Telescope slew replaces the flat `60.0` fudge.
- **`PointingScript` / `AutoFocusScript`**: telescope slew replaces their flat `60.0` fudges.
- **`DarkBiasScript`**: camera's matching `BinningOption.readout_time_s` replaces the flat `5.0`s readout fudge.
- **`SelectorScript`**: unchanged, per the plan's Non-goals (unclear which capability field, if any, actually fits "mode change" duration).

New `TelescopeCapability.estimate_slew_time_s()` (`pyobs/robotic/instruments.py`) centralizes the "typical slew distance / real rate" placeholder shared by all three telescope-slewing scripts, instead of duplicating the constant and the None-rate fallback three times.

## Not in this PR
§A.4 (`PortalTaskArchive`'s concrete fetch/cache) — needs a new pyobs-portal marker endpoint that doesn't exist yet, tracked as its own follow-up. Every backend still returns `None` for `get_instrument_capabilities()` today, so **this PR changes no runtime behavior** until real capability data is actually wired up to reach these scripts.

## Test plan
- [x] 20 new tests across `test_instruments.py` (`TelescopeCapability.estimate_slew_time_s()`) and all 4 scripts' test files (real-value path, missing-module fallback, missing-field fallback, filter-change transition counting including the "None never counts as a change" and cross-repeat-boundary cases)
- [x] `pytest tests/` (excluding `integration`/`xmpp`) — 1886 passed (up from 1866), no regressions
- [x] `ruff check` / `black --check` / `pyrefly check` — clean (30 pre-existing, unrelated pyrefly errors on 2 test files confirmed identical on `develop`)

### DIFF
--- a/pyobs/robotic/instruments.py
+++ b/pyobs/robotic/instruments.py
@@ -70,7 +70,13 @@ class CameraCapability(BaseModel):
 # destination-minus-start-position distance a real estimate would need. See
 # specs/plans/2026-09-01-instrument-capability-duration-estimates.md's "Representative slew/rotate
 # distance" note; pyobs-core#858/#859 track replacing this with a real distance.
-_TYPICAL_SLEW_DEG = 90.0
+#
+# A default, not a hardcoded constant inside estimate_slew_time_s() below: #858/#859's real
+# distances will likely differ per script (a flat-field PointingScript slew and an
+# ImagingScript/AutoFocusScript target slew aren't the same kind of move), and this keeps that
+# future per-caller distance a parameter on the estimate rather than a policy decision baked into
+# what's meant to stay a pure mirror of the portal's data.
+DEFAULT_SLEW_DISTANCE_DEG = 90.0
 
 
 class TelescopeCapability(BaseModel):
@@ -83,14 +89,18 @@ class TelescopeCapability(BaseModel):
     slew_rate_deg_per_s: float | None = None
     updated_at: str | None = None
 
-    def estimate_slew_time_s(self) -> float | None:
-        """First-pass slew-duration estimate: `_TYPICAL_SLEW_DEG` divided by the real slew rate.
-        None if `slew_rate_deg_per_s` isn't declared (or isn't positive) -- callers fall back to
-        their own flat constant in that case, same as when no capability data exists at all.
+    def estimate_slew_time_s(self, distance_deg: float = DEFAULT_SLEW_DISTANCE_DEG) -> float | None:
+        """First-pass slew-duration estimate: `distance_deg` divided by the real slew rate. None
+        if `slew_rate_deg_per_s` isn't declared (or isn't positive) -- callers fall back to their
+        own flat constant in that case, same as when no capability data exists at all.
+
+        `distance_deg` defaults to `DEFAULT_SLEW_DISTANCE_DEG`, the same first-pass placeholder
+        every caller uses today; a caller can pass its own once #858/#859 give it a real,
+        script-specific distance instead of this shared guess.
         """
         if self.slew_rate_deg_per_s is None or self.slew_rate_deg_per_s <= 0:
             return None
-        return _TYPICAL_SLEW_DEG / self.slew_rate_deg_per_s
+        return distance_deg / self.slew_rate_deg_per_s
 
 
 class DomeCapability(BaseModel):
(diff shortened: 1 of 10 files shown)


thusser (commented):
## Review: approve — the §A.8 consumption layer is well-built

Every fallback preserves today's exact formulas, the capability lookups match the model layer's semantics, and the new filter-transition logic is careful and well-tested. No blockers.

### Verified

- **Fallback equivalence holds.** With no capabilities (the production state today — every backend still returns `None`), all four scripts reproduce their old formulas exactly: imaging's readout → 0.0 (it was never counted), slew → flat 60.0, acquisition → +30.0, no filter time; darkbias's `readout = 5.0` in all three branches (exptimes list, match-science, single-exptime); pointing/autofocus → 60.0. So the "no runtime behavior change until data is wired" claim is accurate.
- **Lookup semantics are consistent.** Camera/telescope/filter-wheel lookups key on the scripts' existing module-name fields; readout only applies when the exact `(x, y)` binning is declared on the portal row (conservative — no nearest-binning guessing); `_readout_time_s` returns 0.0 (not a fudge) for imaging since imaging previously counted no readout, while darkbias keeps its 5.0 — matching the plan's per-script fallback semantics.
- **Filter-change counting is correct.** Transitions are counted per config pair (not per exposure — a `count`-long series shares one filter), the sequence wraps across the repeat boundary, `None` (use-current-filter) never counts, and the unknown starting filter isn't counted. Test cases cover R,V,V,R→2, R,V|R,V→3 across repeats, and None-touching→0.
- **`estimate_slew_time_s()`** guards `None`/non-positive rates, and the 90° placeholder is documented with plan + #858/#859 references.
- **Tooling:** 66/66 tests on the 5 touched files pass, full `tests/robotic/scripts/` 139/139 pass, `ruff`/`black` clean. pyrefly's 30 errors are all in `test_autofocus.py`/`test_darkbias.py` on pre-existing lines — I verified the identical error set exists on `develop` (exact line-shift mapping: uniform +1/+2 from the new imports, +27 past the appended test blocks). The "20 new tests" claim checks out (3+2+9+3+3).

### Notes (non-blocking)

1. **Design: estimation policy now lives on a data-mirror model.** `_TYPICAL_SLEW_DEG` + `estimate_slew_time_s()` put a pyobs-side planning decision (the placeholder 90° distance) on `TelescopeCapability`, which #864 deliberately kept a pure serializer mirror. It's self-contained and convenient, but when #858/#859 replace the placeholder with real distances — which will likely differ per script (pointing vs. autofocus slews) — a shared helper parameterized by distance or per-script constants would keep the model clean.
2. **Imaging: the slew term applies regardless of whether the config actually slews.** A pure bias/dark `instrument_configs` sequence still adds slew time (rate-based now, 60.0 flat before) — pre-existing conceptual fudge (the +60 was unconditional), just noting the estimate now distinguishes *rate* but not *whether a slew happens* at all. Also, each filter transition costs one `filter_change_time_s` (portal's documented one-position-step estimate) regardless of how many wheel positions a change actually spans. Both are reasonable first passes per the plan; worth a sentence in the docstring.
3. **Nit:** imaging.py's new `from pyobs.robotic.instruments import CameraCapability` is a runtime import used only in an annotation — the file's own convention (`TaskData`) is `TYPE_CHECKING`; moving it would be consistent (negligible today since `instruments.py` is import-light).

### Summary

Faithful to §A.8, careful fallback design, and the strongest test coverage of the three PRs in this series. Merge as-is; note 1 is the one to keep in mind for the #858/#859 follow-ups.

PR #865, opened 2026-09-02
feat: thread instrument_capabilities through TaskData/TaskArchive/scheduler (§A.2-A.7)

thusser:
## Summary
§A.2-A.7 of `specs/plans/2026-09-01-instrument-capability-duration-estimates.md` — the plumbing that carries an `InstrumentCapabilities | None` (from pyobs-core#864) from the one real caller down to `Task.estimate_duration()`, with no consumer yet and every backend still returning `None`, so **this changes no runtime behavior**.

- `TaskData.instrument_capabilities: InstrumentCapabilities | None` (`task.py`)
- `TaskArchive.get_instrument_capabilities() -> None` by default; every non-portal backend (filesystem, memory, lco) inherits it unchanged
- `Task.estimate_duration()` gains the parameter, forwarded into `TaskData`
- `TaskScheduler.schedule()` gains it in the abstract signature. `AstroplanScheduler` accepts it for interface consistency but doesn't use it (reads the stored `Task.duration` field, not a live estimate — see the plan's Non-goals)
- `OnDemandScheduler` threads it through the full call chain: `schedule` → `schedule_in_interval` → `schedule_first_in_interval` → `find_next_best_task`/`check_for_better_task`/`can_postpone_task`/`create_scheduled_task` → the 4 `task.estimate_duration()` call sites
- `pyobs/modules/robotic/scheduler.py`'s `_schedule_worker` — the only real caller of `schedule()` outside tests — now supplies `self._task_archive.get_instrument_capabilities()`

## Not in this PR
- §A.4: `PortalTaskArchive`'s concrete fetch/cache override (still returns `None` via the base class default)
- §A.8: the 5 leaf scripts actually reading `instrument_capabilities`

Both tracked in the plan doc as follow-up work.

## Test plan
- [x] New tests: `TaskData`/`Task.estimate_duration()` forwarding, `TaskArchive` default, `OnDemandScheduler.create_scheduled_task()` + full `schedule()` end-to-end threading, `pyobs/modules/robotic/scheduler.py`'s `_schedule_worker` forwarding `get_instrument_capabilities()`
- [x] `pytest tests/` (excluding `integration`/`xmpp`) — 1866 passed, no regressions
- [x] `ruff check` / `black --check` / `pyrefly check` — clean (pre-existing, unrelated pyrefly noise on 2 test files confirmed identical on `develop`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019XoNCe7YyJELc8xup1zVdE

PR #868, opened 2026-09-03
feat: PortalTaskArchive polls and caches instrument capabilities (§A.4)

thusser:
## Summary
`PortalTaskArchive.get_instrument_capabilities()` now actually returns live data instead of inheriting `TaskArchive`'s `None` default — the last piece of `specs/plans/2026-09-01-instrument-capability-duration-estimates.md`'s §A.

- New `_poll_instrument_capabilities()`, called from the existing `_poll()` alongside the tasks/projects check: re-downloads `GET /api/instruments/` only when pyobs-portal#144's `last_instrument_update` marker moves, using the same marker-gated pattern `_poll()`/`_update()` already use for tasks/projects (pyobs-portal#84).
- Kept deliberately independent of the tasks/projects poll: a failure fetching or parsing instrument capabilities is caught and throttled locally, keeps serving the last-good `InstrumentCapabilities` rather than clearing it, and never blocks or forces a retry of the tasks/projects side.

## Resolves both open review notes from #864/#867
- **`extra="forbid"` conflict**: a portal payload the pyobs-core models don't recognize now raises inside `_poll_instrument_capabilities()`'s own try/except, degrading like any other fetch failure rather than propagating.
- **Pagination**: reuses the existing `http_request_paginated` helper already used for tasks/projects, so no truncation risk on a fleet with >100 instruments.

## Status of the plan
§A is now fully landed end-to-end in pyobs-core (#864, #865, #867, this PR). Real capability data only actually reaches the scheduler once pyobs-portal#144 (merged) and this PR are both deployed together with a `PortalTaskArchive`-backed scheduler module — and pyobs-portal's own `pyobs-core` dependency needs bumping to a release containing #864-867 first (see #144's review).

## Test plan
- [x] 8 new tests: first-poll download, marker-unchanged skip, marker-moved re-download, last-good-cache-kept on marker-fetch failure / download failure / unparseable payload, `get_instrument_capabilities()` default, `_poll()` actually calls the new method
- [x] `pytest tests/robotic/storage/portal/` — 59/59 passing (51 existing + 8 new)
- [x] `pytest tests/` (excluding `integration`/`xmpp`) — 1895 passed, no regressions

### DIFF
--- a/pyobs/robotic/storage/portal/taskarchive.py
+++ b/pyobs/robotic/storage/portal/taskarchive.py
@@ -100,12 +100,17 @@ class PortalTaskArchive(TaskArchive):
         the tasks/projects poll, and keeps serving the last-good ``InstrumentCapabilities`` rather
         than clearing it -- the same "optional/degrade to None everywhere, never raise" convention
         as the rest of this plan (see ``get_instrument_capabilities()``'s own ``None`` default).
+
+        Compares with ``!=``, not ``>``: deleting the row that held the current ``max(updated_at)``
+        moves the marker *backward*, and a strict ``>`` would silently miss that (a removed device
+        would linger in the cache until some later edit happened to push the marker forward again).
+        ``!=`` catches a backward move too, at no extra cost.
         """
         try:
             last_instrument_update = await self._last_instrument_update_time()
             if (
                 self._instrument_capabilities_marker is None
-                or last_instrument_update > self._instrument_capabilities_marker
+                or last_instrument_update != self._instrument_capabilities_marker
             ):
                 data = await http_request_paginated(self._session, urljoin(self._url, "/api/instruments/"), strict=True)
                 self._instrument_capabilities = InstrumentCapabilities.from_api_response(data)
--- a/pyobs/robotic/storage/portal/taskarchive.py
+++ b/pyobs/robotic/storage/portal/taskarchive.py
@@ -5,6 +5,7 @@ from urllib.parse import urljoin
 
 import aiohttp
 
+from pyobs.robotic.instruments import InstrumentCapabilities
 from pyobs.robotic.storage.taskarchive import TaskArchive
 from pyobs.robotic.task import Project, Task
 from pyobs.utils.http import LogThrottle, http_request_paginated, http_request_with_retries
@@ -31,6 +32,8 @@ class PortalTaskArchive(TaskArchive):
         self._last_marker: Time | None = None
         self._projects: list[Project] = list()
         self._tasks: list[Task] = list()
+        self._instrument_capabilities: InstrumentCapabilities | None = None
+        self._instrument_capabilities_marker: Time | None = None
         # First failure logs immediately at ERROR (someone should know right away); repeats
         # during the same outage are throttled to at most one ERROR per minute.
         self._poll_error_throttle = LogThrottle(quiet_for=0.0, interval=60.0)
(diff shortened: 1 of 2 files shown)


thusser (commented):
## Review: approve — clean, faithful §A.4 implementation

This genuinely resolves both review notes raised on #864/#867 (the `extra="forbid"` hard-fail risk and the >100-instrument pagination truncation), with regression tests for each. No blockers.

### Verified

- **Both #864/#867 review notes are resolved in code + tests.** The unparseable-payload regression test (`[{"some_unrecognized_future_field": 1}]`) proves an `extra="forbid"` rejection degrades inside the poll's own try/except — last-good cache kept, marker not advanced — rather than propagating. Pagination uses `http_request_paginated(..., strict=True)`, the same all-pages + no-silent-truncation helper as tasks/projects.
- **Failure-handling details are correct.** `except Exception` doesn't swallow `CancelledError` (BaseException); the throttle `clear("instrument_capabilities")` fires on any successful marker check (even when no download is needed), so a recovered streak resets; the marker advances *only* after a successful parse, so a mid-way failure retries next poll; `_instrument_capabilities` is never cleared on failure — the last-good object keeps being served.
- **Consistent with the existing pattern.** `_poll_instrument_capabilities` mirrors `_poll`'s marker-gating and `_check_for_changes`' `LogThrottle` key/clear idiom; naming (`_instrument_capabilities`/`_instrument_capabilities_marker`) mirrors `_tasks`/`_last_marker`; the `get_instrument_capabilities()` override matches the base's `InstrumentCapabilities | None` signature (pyrefly confirms).
- **Contract matches the merged portal side.** The URL `/api/instruments/last_instrument_update/` and the `Time(res["last_instrument_update"])` epoch-safe parse match pyobs-portal#144 as merged (confirmed `MERGED` — the status claim in the PR body is accurate).
- **Tooling/tests:** 59/59 portal-archive tests pass (51 existing + 8 new, all well-targeted: first-poll download, unchanged-marker skip, moved-marker re-download, three keep-last-good failure modes, and the `_poll()` wiring test); ruff/black/pyrefly clean (0 errors). Full suite: 1888 passed here with only the 7 known environmental failures (sunpy config dir + `/opt/pyobs` read-only in this sandbox), consistent with the claimed 1895 in a healthy environment.

### Notes (minor, non-blocking)

1. **Strictly-greater marker comparison misses backward marker moves.** If the admin *deletes* the row holding the current `max(updated_at)`, the marker decreases and this poll never notices — the cached capabilities keep the removed device until some later edit pushes the marker forward. This exactly mirrors the pre-existing tasks-marker semantics (and the impact is estimate-only data that self-heals on the next edit), so it's fine — but `!=` (with the existing `None` first-poll check) would also catch deletions at zero cost if you want instruments to be deletion-aware where tasks aren't.
2. **Poll ordering coupling (asymmetric).** `_poll_instrument_capabilities()` runs after the tasks block, so a tasks-side exception skips the instrument attempt for that cycle. In practice both fail together (portal down), and the docstring's promise — instrument failures never block/retry the tasks side — holds. Just noting the reverse direction isn't also isolated.
3. **Persistent unparseable payload = full re-download every 5s.** When the marker has advanced but parsing keeps failing (e.g. the portal deployed a new serializer field before this side catches up), each `_poll` cycle re-downloads the entire instrument list and fails again — same cadence and no backoff beyond `http_request_with_retries`' internal retries and the log throttle. Matches the tasks-side pattern; acceptable, but a short failure backoff would be a nice future hardening.

### Summary

This completes §A end-to-end (models → plumbing → leaf scripts → portal fetch), the degradation semantics are exactly what the plan's "optional/degrade-to-None" convention calls for, and the two open review notes from the earlier PRs in this series are closed with genuine regression tests. Merge as-is.

Issue #871, opened 2026-09-03
Comm: unregister_event() doesn't cancel already-scheduled handler tasks (stale-widget race)


thusser:
## Problem

`Comm._send_event_to_module()` dispatches an event to each registered handler by calling it to get a coroutine, then fire-and-forgets it via `asyncio.create_task(ret)`:

```python
# comm.py
if event.__class__ in self._event_handlers:
    for handler in self._event_handlers[event.__class__]:
        ret = handler(event, from_client)
        if asyncio.iscoroutine(ret):
            task = asyncio.create_task(ret)
            task.add_done_callback(functools.partial(self._log_handler_exception, handler=handler, event=event))
```

`create_task()` only *schedules* the coroutine; it doesn't run it inline. If the handler's owner (e.g. a pyobs-gui widget) is torn down and calls `unregister_event()` in the same synchronous stretch before the scheduled task gets a turn, `unregister_event()` removes the handler from `self._event_handlers` but has no way to retract the task that was already created for it. The task still runs later, against whatever the handler is bound to — which for a Qt widget may already have had its underlying C++ object destroyed by then.

## Observed symptom (pyobs-gui)

```
RuntimeError: libshiboken: Internal C++ object (PySide6.QtWidgets.QCheckBox) already deleted.
```

raised from `DataDisplayWidget._on_new_data` (a `NewImageEvent`/`NewSpectrumEvent` handler) touching `self.checkAutoUpdate` after the widget was discarded (e.g. on client disconnect). `_log_handler_exception` already catches and logs this, so it isn't fatal, but it's log noise masking a real dispatch-ordering bug, and the same race applies to any handler bound to an object that gets destroyed on teardown (pyobs-gui already has to work around it elsewhere with ad-hoc `discard()`/unsubscribe tracking, e.g. `StatusItem._state_subscriptions`, `MainWindow.discard_all_widgets()`).

## Proposed fix

Have `Comm` track in-flight dispatch tasks per `(event_class, handler)`, and have `unregister_event()` cancel any tasks that are still pending for that handler at unregistration time. This closes the race at the source for every consumer, instead of requiring each handler to defensively guard against operating on a torn-down target.

## Where

`pyobs/comm/comm.py`: `_send_event_to_module()` (dispatch/task creation) and `unregister_event()`.

thusser:
[...]

PR #876, opened 2026-09-04
fix: Comm.unregister_event() cancels already-scheduled handler tasks

thusser:
## Summary

- `Comm._send_event_to_module()` fire-and-forgets each handler dispatch via `asyncio.create_task()`, which only schedules the coroutine — it doesn't run inline.
- If a handler's owner tears down and calls `unregister_event()` in the same synchronous stretch, before the scheduled task gets a turn, the task ran anyway against a possibly torn-down target (observed in pyobs-gui as `RuntimeError: libshiboken: Internal C++ object already deleted`).
- Tracks in-flight dispatch tasks per `(event_class, handler)` and cancels any still pending for a handler `unregister_event()` actually removes, closing the race at the dispatch layer.

Plan: `specs/plans/2026-09-03-comm-unregister-event-task-cancellation.md`

Closes #871

## Test plan

- [x] New regression tests in `tests/comm/test_events.py`: task scheduled then unregistered before the event loop yields gets cancelled and its body never runs; a handler's pending task for a *different* event type is unaffected.
- [x] `uv run pytest tests/comm/ -m "not integration and not xmpp"` — 182 passed
- [x] `uv run pytest -m "not integration and not xmpp"` (full suite) — 1912 passed
- [x] `uv run ruff check`, `uv run black --check`, `uv run pyrefly check` — clean

https://claude.ai/code/session_019SjkJ4CSYmbqB3fhef5JwJ

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ v2.5.0
   ``HIERARCH <MODULE> VERSION <PACKAGE>`` card per loaded ``pyobs-*`` package into FITS headers,
   reusing ``loaded_pyobs_packages()`` (added for #759). Closes #739. See
   ``specs/design/package_versions_fits_header.md``.
+* Fixes ``Comm.unregister_event()`` not cancelling a handler dispatch that
+  ``_send_event_to_module()`` had already scheduled via ``asyncio.create_task()`` before the
+  unregister call -- the task ran later regardless, against whatever the handler was bound to
+  (observed in pyobs-gui as a ``libshiboken: Internal C++ object already deleted`` error on
+  client disconnect). Closes #871.
 
 v2.4.1 (2026-09-03)
 *********************
--- a/pyobs/comm/comm.py
+++ b/pyobs/comm/comm.py
@@ -39,6 +39,7 @@ class Comm:
         self._log_queue: asyncio.Queue[LogEvent] = asyncio.Queue()
         self._logging_task: asyncio.Task[Any] | None = None
         self._event_handlers: dict[type[Event], list[Callable[[Event, str], Coroutine[Any, Any, bool]]]] = {}
+        self._event_handler_tasks: dict[tuple[type[Event], Callable[..., Any]], set[asyncio.Task[Any]]] = {}
         self._events_sent: set[type[Event]] = set()
         self._events_subscribed: set[type[Event]] = set()
         self._closing = asyncio.Event()
@@ -479,6 +480,15 @@ class Comm:
                     self._events_subscribed.discard(ev)
                     unsubscribed.append(ev)
 
+                # cancel any dispatch for this handler that _send_event_to_module() already
+                # scheduled via asyncio.create_task() but that hasn't run yet -- otherwise it
+                # fires later against whatever the handler is bound to, even though the handler
+                # is gone from _event_handlers as of this call (see issue #871)
+                pending = self._event_handler_tasks.pop((ev, handler), None)
+                if pending is not None:
+                    for task in pending:
+                        task.cancel()
+
         # only event classes that just lost their last handler need the comm layer to actually
         # tear anything down (e.g. XmppComm unsubscribing from peers' event nodes)
         if unsubscribed and not event_class.local:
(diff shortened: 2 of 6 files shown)


thusser (commented):
**Verdict: approve with nits.** The fix is correct, minimal, and targets the actual race described in #871. I reviewed it against `develop` and reproduced the checks locally: `tests/comm/` → 182 passed, `ruff`/`black --check` clean, `pyrefly` 0 errors. CI is green as well.

**Verified in the code**
- Dispatch/tracking key consistency: `_send_event_to_module()` keys tasks by `(event.__class__, handler)`, and `unregister_event()` iterates the same derived-events expansion `register_event()` used, popping per `ev` — so the pop always finds what dispatch recorded, without over-cancelling the same handler's other event types.
- Cancel-before-start semantics hold on the supported Python line: the new test schedules a dispatch and unregisters in the same synchronous stretch and confirms the handler body never runs.
- The remove/pop/cancel loop is synchronous with no awaits (no interleaving), the cancel loop runs before the XMPP-only `_unregister_events` branch (so local and non-local events are both covered), and done-callbacks that fire after a pop no-op via `.get(...) → None` — no KeyError path.
- Bookkeeping hygiene: the done-callback discards tasks and deletes empty sets on both completion and exception paths.
- GUI flow matches the guarantee: pyobs-gui's `BaseWidget.discard()` awaits `comm.unregister_event()` inline, and `discard_all_widgets()` awaits each `widget.discard()` before Qt destroys the widget — so cancellation runs before the pending dispatch task gets its turn in the reported scenario.

**Points to consider (non-blocking)**
1. **Semantic widening is undocumented.** `unregister_event()` previously meant "stop future deliveries"; it now also interrupts an already-started handler suspended at an `await` (it gets `CancelledError`). That's what teardown wants, but the method docstring doesn't mention cancellation at all — worth one sentence stating the new contract.
2. **Guarantee boundary should be stated.** The fix only wins if the cancel loop executes before the event loop next runs the pending task, i.e. `unregister_event()` must be *awaited inline* in the same stretch. A fire-and-forget teardown (`create_task(unregister_event(...))`) queued behind the dispatch task would still lose the race. pyobs-gui satisfies this today, but a docstring sentence would pin the contract for future callers.
3. **Set semantics untested.** The cancel-all loop only ever iterates single-element sets in the tests; a test with two pending tasks under one `(ev, handler)` key would exercise the actual set design.
4. **Memory-hygiene claim untested.** The docstring promises the dict doesn't accumulate because the done-callback removes entries, but nothing asserts a normally-completed handler's key disappears. Given this fix is about widget lifetime, a regression test (dispatch → let the task finish → assert key removed) is worth adding — it's the guard against the dict becoming a new strong-reference path keeping discarded widgets alive.
5. **Test boilerplate.** Every unit test hand-rolls `Comm.__new__(Comm)` plus five attribute assignments, and this PR had to touch all of them. A small `bare_comm()` helper would prevent the next internal-attribute addition from touching every test.
6. **Minor.** The dict annotation uses `Callable[..., Any]` while `_event_handlers` uses the precise handler signature — a shared alias would be tighter. Also, the plan doc's *Problem* section still cites pre-fix line numbers (`comm.py:678-693`, `458-485`); now that it's marked `implemented`, those describe the old layout.
7. **CHANGELOG (optional).** Recent fix PRs (#824, #830) carried an entry; since this closes #871 with a user-visible symptom in pyobs-gui, an entry is defensible if you follow the per-fix convention.

thusser:
Addressed all 7 points:

1. `unregister_event()` docstring now states the cancellation contract, including that it can raise `CancelledError` into an already-suspended handler.
2. Same docstring update also states the inline-await requirement (a `create_task(unregister_event(...))` caller can still lose the race).
3. New test: two events dispatched to the same `(event_class, handler)` before unregistering — both cancelled, exercising the set with >1 element.
4. New test: a dispatch that completes normally (not cancelled) gets its `(event_class, handler)` entry removed from `_event_handler_tasks`.
5. Added a `_bare_comm()` helper in `tests/comm/test_events.py`; all existing tests in that file now use it instead of the 4-line boilerplate.
6. Added an `EventHandler` type alias and used it everywhere the inline `Callable[[Event, str], Coroutine[Any, Any, bool]]` signature was repeated, including the new tracking dict. Plan doc's Problem section now marks its line numbers as pre-fix.
7. Added a CHANGELOG entry under the current unreleased heading, matching the #824/#830 convention.

Full non-integration suite (1914 tests), ruff, black, pyrefly all clean on the updated branch.

PR #882, opened 2026-09-04
Add camera/filter-wheel model fields, fix nullable module_name bug

thusser:
## Summary
- `CameraCapability` gains `model` (e.g. "FLI ProLine PL23042") and `sensor_type` (e.g. "e2v CCD230-42, back-illuminated CCD"); `FilterWheelCapability` gains `model` (e.g. "FLI CFW-2-7"). Pure reference data, no behavior change.
- `FilterWheelCapability.module_name`: nullable → required. It was documented as "for a wheel with no addressable module of its own", but `InstrumentCapabilities.__init__` only ever indexes non-null module names — a null row was **silently unreachable** by any script/scheduler lookup. The "integrated into the camera" case it was meant for is correctly handled by entering the *camera's* own module_name instead, not leaving it blank.
- Every model in `pyobs/robotic/instruments.py` now inherits a new `_ForwardCompatibleModel` base (`extra="ignore"` instead of `BaseModel`'s `extra="forbid"`). Discovered live: adding the two fields above to pyobs-portal immediately broke a running `mastermind` still on an older pyobs-core release — it degraded gracefully (kept serving the last-good cache, per existing design) but logged a parse error on every poll until upgraded. Scoped narrowly to this module (portal-controlled data, not user-authored config) — `extra="forbid"` stays the right default everywhere else in pyobs.

## Why
Surfaced while filling in MONET-South's real instrument capability data (`monet/pyobs-monet#3`) through the portal admin — wanted to record the actual camera/sensor model, hit the nullable-module_name bug while discussing why filter wheels have a `name` field and cameras don't, and hit the forward-compat gap the moment the new fields landed on a live portal.

Companion PR on the portal side: pyobs-portal side (same title).

## Test plan
- [x] `pytest tests/robotic/` — 504 passed
- [x] Rewrote `test_instrument_capabilities_poll_keeps_last_good_on_unparseable_payload` (tested the old `extra="forbid"` behavior directly) into `test_instrument_capabilities_poll_tolerates_unrecognized_fields`, verifying the new degrade-gracefully-and-still-update behavior

### DIFF
--- a/pyobs/robotic/instruments.py
+++ b/pyobs/robotic/instruments.py
@@ -10,12 +10,29 @@ from __future__ import annotations
 
 from typing import Any
 
-from pydantic import Field
+from pydantic import ConfigDict, Field
 
 from pyobs.utils.serialization import BaseModel
 
 
-class Filter(BaseModel):
+class _ForwardCompatibleModel(BaseModel):
+    """Base for the capability models below, relaxing `BaseModel`'s `extra="forbid"` to
+    `extra="ignore"`.
+
+    A running process (e.g. `mastermind`) can be on an older pyobs-core release than whatever
+    portal it polls -- the portal gaining a field (as it will over time; `model`/`sensor_type`
+    already did once) must not turn into a hard parse failure that starves it of every other
+    field in the response. `extra="forbid"` is a deliberate, useful default elsewhere in pyobs
+    (catching config typos, `2026-08-15-pydantic-extra-validation.md`) but wrong here: this
+    module only ever consumes portal-controlled data, not user-authored config, and the two
+    sides (portal, and every fleet site's own pyobs-core version) are never guaranteed to be on
+    the same release.
+    """
+
+    model_config = ConfigDict(extra="ignore")
+
+
+class Filter(_ForwardCompatibleModel):
     """A single filter position in a filter wheel."""
 
     name: str
@@ -23,7 +40,7 @@ class Filter(BaseModel):
     updated_at: str | None = None
 
 
-class BinningOption(BaseModel):
+class BinningOption(_ForwardCompatibleModel):
     """Readout time for one binning mode of a camera."""
 
     x: int
@@ -32,25 +49,31 @@ class BinningOption(BaseModel):
     updated_at: str | None = None
 
 
-class FilterWheelCapability(BaseModel):
-    """Planning-time capability data for one filter wheel.
+class FilterWheelCapability(_ForwardCompatibleModel):
+    """Planning-time capability data for one filter wheel, keyed by `module_name`.
 
-    `module_name` is nullable -- not every filter wheel is its own addressable module (some
-    cameras expose filter selection through the camera module itself), per pyobs-portal#142.
+    `module_name` is required, same as every other device here -- a wheel with filter selection
+    exposed through the camera's own module (no independent XMPP identity) should be entered with
+    the *camera's* module_name, not left blank. It used to be nullable (pyobs-portal#142) but a
+    blank value made the row permanently unreachable via `InstrumentCapabilities.filter_wheel()`,
+    which indexes by module_name -- there was no valid case for it in practice.
     """
 
     name: str = ""
-    module_name: str | None = None
+    module_name: str
+    model: str = ""
     filter_change_time_s: float | None = None
     updated_at: str | None = None
     filters: list[Filter] = Field(default_factory=list)
 
 
-class CameraCapability(BaseModel):
+class CameraCapability(_ForwardCompatibleModel):
     """Planning-time capability data for one camera, keyed by `module_name`."""
 
     module_name: str
     code: str
+    model: str = ""
+    sensor_type: str = ""
     pixel_size_um: float | None = None
     sensor_width_px: int | None = None
     sensor_height_px: int | None = None
@@ -79,7 +102,7 @@ class CameraCapability(BaseModel):
 DEFAULT_SLEW_DISTANCE_DEG = 90.0
 
 
-class TelescopeCapability(BaseModel):
+class TelescopeCapability(_ForwardCompatibleModel):
     """Planning-time capability data for one telescope, keyed by `module_name`."""
 
     module_name: str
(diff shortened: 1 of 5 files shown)


tests/robotic/storage/portal/test_portal_archives.py
```diff
@@ -996,21 +996,28 @@ async def test_instrument_capabilities_poll_keeps_last_good_on_download_failure(
...
-        AsyncMock(return_value=[{"some_unrecognized_future_field": 1}]),
+        AsyncMock(return_value=[{"display_name": "Fresh", "some_unrecognized_future_field": 1, "cameras": []}]),
     )
     await archive._poll_instrument_capabilities()
-    assert archive.get_instrument_capabilities() is good
-    assert archive._instrument_capabilities_marker == T1
+
+    capabilities = archive.get_instrument_capabilities()
+    assert capabilities is not stale
+    assert isinstance(capabilities, InstrumentCapabilities)
```
thusser, line 1018: This rewrite drops the only test that pinned the *parse-failure* branch: `ValidationError` out of `from_api_response` → keep last-good, marker NOT advanced. After this change an unrecognized field can no longer fail parsing, so the remaining reasons `from_api_response` raises are structural (missing required `module_name`/`code`, wrong type, a portal *rename* of a required field) — and the except-branch/marker behavior for exactly those is now only covered indirectly by the ConnectionError tests. Suggest a companion test feeding a structurally invalid payload (e.g. a camera missing required `module_name`/`code`) and asserting last-good is kept and the marker stays at T1 (so the next poll retries). That is the safety net for the next portal drift that *isn't* ignorable.

pyobs/robotic/instruments.py
```diff
@@ -32,25 +49,31 @@ class BinningOption(BaseModel):
...
+    `module_name` is required, same as every other device here -- a wheel with filter selection
+    exposed through the camera's own module (no independent XMPP identity) should be entered with
+    the *camera's* module_name, not left blank. It used to be nullable (pyobs-portal#142) but a
+    blank value made the row permanently unreachable via `InstrumentCapabilities.filter_wheel()`,
+    which indexes by module_name -- there was no valid case for it in practice.
     """
 
     name: str = ""
-    module_name: str | None = None
+    module_name: str
```
thusser: `module_name: str` makes the field required but still accepts `""` — pydantic here enforces not-None, not non-empty (verified: `FilterWheelCapability(name="x", module_name="")` parses fine and lands in the index under `""`). The portal side's `blank=False` stops new empty values and the #152 data check confirmed no NULL rows, but #142-era rows could hold `""` (`blank=True` allowed it), which the migration's NOT NULL change would not catch and which would be silently unreachable — the same dead-data bug in a new disguise. Cheap hardening: `Field(min_length=1)` here, and extend the portal data check to empty strings.

thusser (commented):
Review pass over pyobs-core#882 (mirrored portal change in pyobs-portal#152). Verified locally on this branch: `tests/robotic/test_instruments.py` + `tests/robotic/storage/portal/test_portal_archives.py` (82 passed), plus direct behavior probes — missing `module_name` now raises `ValidationError`, and unknown fields are tolerated at every nesting level (Instrument/Camera/FilterWheel/Filter/Telescope/Dome/Roof).

Overall this is well-scoped and the write-up is accurate: the §A.1 nullable-`module_name` row really was unreachable dead data (`InstrumentCapabilities.__init__` only indexes non-None), and `extra="ignore"` cleanly resolves the open question the 2026-09-01 duration-estimates plan explicitly flagged (§4 "extra='forbid' on the §A.1 models degrade-to-None conflict") — choosing option (b) with the reasoning captured in the `_ForwardCompatibleModel` docstring. Two non-blocking things worth addressing:

1. **Policy consistency** (no code change needed, but worth a sentence): `specs/plans/2026-08-15-pydantic-extra-validation.md` explicitly decided *self-hosted portal* payloads stay strict — "declare the missing fields, use `forbid` here too, no carve-out" — and the `Task`/`Project` models follow that convention today (`updated_at`, `public` were each added to the strict models in lockstep, with tests literally named `..._accepts_updated_at` / `..._accepts_public`). This PR is a deliberate carve-out for the same "portal ahead of core" scenario, and the justification in the docstring ("this module only ever consumes portal-controlled data, not user-authored config") applies equally to `Task`/`Project`, which also parse the same portal and also degrade-to-last-good on failure — the identical incident will recur the next time `/api/tasks/` or `/api/projects/` grows a field before the core mirror lands. I agree the carve-out is right for capabilities (open-ended reference-data field set vs. curated task/project payloads that get re-validated into config-like shapes), but the plan doc should say out loud that `Task`/`Project` are *not* covered and why, so the next field addition doesn't rediscover this the way the mastermind incident did.

2. **Two small robustness gaps**, commented inline: the parse-failure branch of the poll lost its only direct test, and `module_name: str` still accepts `""` (pydantic enforces not-None, not non-empty).

Minor: `test_instrument_tolerates_unknown_fields` only injects unknown fields at Instrument/Camera/Telescope level — the shared base makes uniform behavior likely, but a couple of extra levels (FilterWheel/Filter/Binning) would make the property explicit rather than implied.

thusser:
Addressed all points from the review:

1. **`module_name` empty-string gap** — `Field(min_length=1)` added, broadened to every capability model's `module_name` (not just `FilterWheelCapability`'s), since the same unreachable-row bug applies equally everywhere. New test: `test_module_name_min_length_applies_to_every_capability_model`.
2. **Lost structural-failure coverage** — added `test_instrument_capabilities_poll_keeps_last_good_on_structurally_invalid_payload` (a camera missing required `module_name`/`code` — real `ValidationError`, not something `extra="ignore"` touches — still degrades to last-good, marker not advanced).
3. **Task/Project policy consistency** — added an explicit paragraph to the plan doc's §3 stating Task/Project deliberately keep `extra="forbid"` per `2026-08-15-pydantic-extra-validation.md`, why the capability models are different (open-ended reference data vs. curated config-like payloads), and that the fix if this trade-off stops holding is "keep Task/Project in lockstep," not relax them too.
4. **Nested unknown-fields coverage** — extended `test_instrument_tolerates_unknown_fields` to cover Camera/Binning/FilterWheel/Filter/Telescope/Dome, not just the top three levels.
5. Added the deploy-ordering constraint (core first) explicitly to the plan doc as its own section.

507 tests pass, ruff/black/pyrefly clean.

